Issue #84, opened 2019-02-15
Add support for WOD Timewalking (Infinite Timereaver)


SacredDuckwhale:
https://www.wowhead.com/news=290247/warlords-of-draenor-timewalking-preview-dungeons-locations-vendors

SacredDuckwhale:
Added via #107 - with the caveat that there may be some inaccuries:
> I have tested this and have not found a boss where the counting does not work. There is one case that could go wrong which would probably require some core code. On the iron docks there is a fight with 3 bosses. If you kill the main one and wipe it would have counted that one. I couldn't find a mechanism in your addon to require multiple bosses for a count.

SacredDuckwhale:
Added in e92a547cff5d78f40419e261bcebc927625a231a.

PR #107, opened 2019-05-30
Added WoD Timewalking instances to Infinite Timereaver list

redpandamonium:
Fixes issue #84

### DIFF
--- a/Options_Defaults.lua
+++ b/Options_Defaults.lua
@@ -99,7 +99,7 @@ R.string_archraces = {
 	[16] = L["Nerubian"],
 	[17] = L["Night Elf"],
 	[18] = L["Fossil"],
-	[19] = L["Draenei"],	
+	[19] = L["Draenei"],
 	[20] = L["Dwarf"],
 }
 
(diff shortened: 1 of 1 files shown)


SacredDuckwhale:
This is very helpful, thanks! Due to the code format changes the PR can't be merged without conflicts, so I've manually added it to the latest alpha.

One more question, though: Aren't there encounters where the boss survives when defeated, such as Yalnu? If so, the detection may not work unless special handling is added for those, usually via statistics (see OnCriteriaComplete() in Core.lua).

I can't test this currently and I'm also not very familiar with all the WOD dungeons, so I might be wrong.

redpandamonium:
I have tested this and have not found a boss where the counting does not work. There is one case that could go wrong which would probably require some core code. On the iron docks there is a fight with 3 bosses. If you kill the main one and wipe it would have counted that one. I couldn't find a mechanism in your addon to require multiple bosses for a count.

SacredDuckwhale:
Excellent. I'll add that part as a potential inaccuracy to the respective issue, but I don't believe adding more code for this fringe case would be reasonable (and in any case I can't test it right now as my account is inactive).

If you happen to run the dungeon with the debug log enabled, or use /rarity dump right after defeating that particular boss, you should see something along the lines of "Detected achievement criteria completion: <id>" - If there's one for this encounter then that could be added as a special case to the OnCriteriaComplete() function I mentioned, so that it'll count an attempt after the encounter was defeated.

SacredDuckwhale:
Added in e92a547cff5d78f40419e261bcebc927625a231a :)

PR #188, opened 2020-07-16
DB: Fix rarity of Armored Vaultbot

borgiman:
Current rarity is 1 in 20 but according to my personal experience and the droprate its much more rare.
1.6%: https://www.wowhead.com/item=170072/armored-vaultbot


SacredDuckwhale:
Thanks! Most drop rates are merely (more or less) educated guesses, so they're often quite inaccurate :)

PR #189, opened 2020-08-01
DB: Correct MapID for Urgloax/Chalice of Secrets

Daeymien:
Urgolax is in Vale of Eternal Blossoms, not inside Shrine within the Vale


SacredDuckwhale:
Thanks! Seems like an oversight :)

PR #193, opened 2020-08-24
Updates for Gahz'rooki's Summoning Stone

domi86:
fixed npcs and creatureId, added coords

Note: According to wowhead comments (https://www.wowhead.com/npc=71159/gahzrooki#comments) this has recently been changed to drop from different mobs


SacredDuckwhale:
Thanks! I guess there's no official confirmation of this change, but it does seem the old NPC is now [something else entirely](https://www.wowhead.com/npc=141789/kill-credit-cave).

domi86:
I couldn't find anything official. But the butchers do drop this item. Found it after about 700 kills

PR #208, opened 2020-10-20
Added option to hide already-known items from the mouseover tooltip.

dbowden713:
Implementing feature for issue/request posted [here](https://github.com/SacredDuckwhale/Rarity/issues/201).

Created a new DB variable `hideKnownItemsInTooltip` for tracking the new option, which is added to the World Tooltip Options. This option is checked before Rarity adds any info to the tooltip where an item could be already known.

Tested as many types of drops as I could, but there may be a few things left to check.

Note: doesn't hide achievement info.

Currently works for: 
- rare drops
- one-time drops
- zone drops
- drops from items (like boxes)

New option:
option


Core/GUI.lua
```diff
@@ -1881,40 +1881,44 @@ _G.GameTooltip:HookScript(
...
-								attemptText = ""
-							end
-							GameTooltip:AddLine(
-								colorize(
-									(not rarityAdded and L["Rarity: "] .. (R.db.profile.blankLineAfterRarity and "\n" or "") or "") ..
-										(itemLink or itemName or v.name) .. attemptText,
-									yellow
+							if v.known and R.db.profile.hideKnownItemsInTooltip then
+								GameTooltip:Show()
+							else
```
dbowden713, line 1886: This is the main change. Everything else was just moved inside the `else` block.

SacredDuckwhale:
Thank you! Could you rebase your branch to get rid of the unrelated merge commit (https://github.com/SacredDuckwhale/Rarity/pull/208/commits/3d51715c9623232d689eac2c3c3ba073b16bf624)?

I tend to use rebase when updating locally to avoid unnecessary merge commits that will persist in history.

I'll take a look at the rest ASAP, but I think simply forking the latest master and applying your changes is the easiest way to get this done.

dbowden713:
Sorry about that! I've squashed everything into one commit and it seems to be cleaned up now. If I need to do anything else on my end just let me know.

SacredDuckwhale:
The only thing that strikes me is the commit description still saying ``Fix blank line appearing in tooltips with already-known items hidden.`` :)

dbowden713:
There we go!

SacredDuckwhale:
Thanks again or your contribution! I'll add this to the latest alpha.

PR #232, opened 2020-12-15
Fix incorrect covenant for Hopecrusher Gargon

nickbayley:
**Hopecrusher Gargon** drops for **Venthyr** and not **Necrolord**. I also noticed that the constant was spelt incorrectly for **Harika the Horrid**.

My editor also stripped some white space. Sorry if that is an issue.

### DIFF
--- a/Options_Defaults.lua
+++ b/Options_Defaults.lua
@@ -1458,7 +1458,7 @@ function R:PrepareDefaults()
 				{ m = CONSTANTS.UIMAPIDS.REVENDRETH, x = 46.0, y = 78.5, n = L["Harika the Horrid"] },
 			},
 			requiresCovenant = true,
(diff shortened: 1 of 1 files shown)


nickbayley:
I see that you fixed Hopecrusher Gargon in this commit: https://github.com/SacredDuckwhale/Rarity/commit/c19805a5d444ea8998a7b0c3c87b9e378991cc05

I didn't realise this was a mirrored repository.

SacredDuckwhale:
It's not a mirrored repository, I just forgot to push the master branch too after pushing the tag. My bad, not used to anyone actually seeing what I do in here :P

The whitespace thing happened to me, as well. It's annoying and shouldn't happen, but the codebase is very messy and not at all standardized, so sometimes there's collateral damage if you aren't very careful. Oh well.

Issue #197, opened 2020-09-25
Add option to only show the addon window on click (instead of on hover)


SacredDuckwhale:
Source: [Curse](https://www.wowace.com/projects/rarity?comment=2250)
> Can you make the minimap icon not work on hover but make it a button. I always end up accidentally activating it by hovering over it.

PR #233, opened 2020-12-18
Add tooltip activation options, requested in SacredDuckwhale/Rarity#197

jelveby:
This adds the option to change tooltip activation to click instead of hover, as requested in #197.

If "On click" is selected instead of "On hover" (which is the default), activating the tracker widget requires a `ctrl + shift + click`, but that is only described in the hoverable description of the option, which isn't necessarily super user friendly.

I couldn't find any rules regarding contributions, so feel free to just use this as reference if there's something else that needs to be done that I'm not privy to.


SacredDuckwhale:
Hello, and many thanks for your contribution! I've tested your changes and found a couple of smaller issues after enabling the "on click" mode.

Firstly, with the LDB plugin:

* When tooltip activation is set to "hover", clicking on the LDB icon immediately closes the tooltip. This is most likely because it isn't rendered at the cursor's position nor updated after moving the cursor, like it would when hovering to toggle it

As well as for the minimap button:

* The button should be a proper toggle, i.e. clicking it twice should close it again to provide a consistent user experience. The normal way to hide the tooltip when moving the cursor outside of the tooltip frame still works, but is less intuitive when using this mode IMHO
* Since the tooltip no longer automatically opens in this mode, there should be some indication as to how it can be opened, e.g., showing a tooltip that says "Click to toggle the tooltip" or similar
* The progress bar can no longer be displayed (CTRL/SHIFT-click simply opens the options for me). I suggest restoring the right-click functionality, so that left-clicking opens the tooltip and right-clicking still toggles the progress bar, which matches the behaviour of the "hover" mode (see [WOW API: IsMouseButtonDown](https://wow.gamepedia.com/API_IsMouseButtonDown))

I'd also want to change some of the names in the code, but that can wait until all functionality is ready (and I could easily do it myself if need be).

There aren't any official contribution guidelines currently, so I'm mostly just going by what comes to mind when testing a new feature or modification :)

jelveby:
Thanks for the feedback!

I agree completely with all the minimap feedback and I've implemented all of the changes.

As for your LDB plugin feedback, I'm not sure exactly how to reproduce your issue since I'm not sure exactly which icon/addon you're referring to. Looking at the recommended addons in the README I tried with both Button Bin and Titan Panel and had no issues that I could see.

Oh, and about naming? I'm gonna be REALLY mad if you change... No, but seriously, I know how particular I can be with naming in my own projects, so you go right ahead. When it comes to other people's projects, they're the boss. 😉

jelveby:
I just noticed that there's still some inconsistency with clicking and shift-clicking. Pretty sure I just need to look a bit closer at the ordering of the if statements. I'll have that fixed soon.

jelveby:
Those inconsistencies should now be fixed.

SacredDuckwhale:
That's really cool; I love how it works now :) There's just a few minor details to resolve, on which I'll comment separately.

Core/GUI.lua
```diff
@@ -1589,8 +1598,64 @@ local function addGroup(group, requiresGroup)
...
+		-- intentionally one column more than we need to avoid text clipping
+		quicktip:SetScale(self.db.profile.tooltipScale or 1)
+	end
+
+	quicktip:AddHeader('Rarity')
+	quicktip:AddSeparator(1, 1, 1, 1, 1)
+	quicktip:AddLine('Left click', 'Open Rarity window')
+	quicktip:AddLine('Right click', 'Toggle tracker')
+	quicktip:AddLine('Shift + Left click', 'Open settings')
+	quicktip:AddLine('Ctrl + Left click', 'Change sorting')
```
SacredDuckwhale, line 1631: These displayed lines should probably be localized, as well

DB/SharedConstants.lua
```diff
@@ -7,6 +7,12 @@ addonTable.constants = addonTable.constants or {}
 
 local C = addonTable.constants
 
+-- Types of items
+C.TOOLTIP = {
+	ACTIVATION_HOVER = "TIP_ACTIVATION_HOVER",
+	ACTIVATION_CLICK = "TIP_ACTIVATION_CLICK",
+}
```
SacredDuckwhale, line 14: I think a slightly more accurate name would be helpful, like ``ACTIVATE_ON_HOVER`` or ``ACTIVATION_METHOD_HOVER``, or similar

Modules/Options/Options.lua
```diff
@@ -47,6 +47,10 @@ local FEED_MINIMAL = "FEED_MINIMAL"
 local FEED_NORMAL = "FEED_NORMAL"
 local FEED_VERBOSE = "FEED_VERBOSE"
 
+-- Tooltip activation
+local TIP_ACTIVATION_HOVER = "TIP_ACTIVATION_HOVER"
+local TIP_ACTIVATION_CLICK = "TIP_ACTIVATION_CLICK"
+
```
SacredDuckwhale, line 53: Shouldn't this be read from the SharedConstants DB? This way it's defined in just one place and not three

Options_Defaults.lua
```diff
@@ -114,6 +114,10 @@ local TIP_LEFT = "TIP_LEFT"
 local TIP_RIGHT = "TIP_RIGHT"
 local TIP_HIDDEN = "TIP_HIDDEN"
 
+-- Tooltip activation
+local TIP_ACTIVATION_HOVER = "TIP_ACTIVATION_HOVER"
+local TIP_ACTIVATION_CLICK = "TIP_ACTIVATION_CLICK"
+
```
SacredDuckwhale, line 120: Already defined in SharedConstants, so we should always refer to that instead of re-defining it :)

SacredDuckwhale (changes requested):

Modules/Options/Options.lua
jelveby: That might be a good idea (if sharing the contstants from the main module is straightforward), but then that should probably be a separate refactor since it looks to me like a bunch of other things are also double defined in the options module.

Options_Defaults.lua
jelveby: I agree with this, but as with my comment on the previous snippet, this also has some previous duplication that might be candidates for refactoring.

Modules/Options/Options.lua
SacredDuckwhale: I figured it's a trivial change so we might as well do it right the first time around. Just because the original author copy/pasted constants everywhere, doesn't mean we have to keep doing it :)

They're exported via ``Rarity.CONSTANTS`` in ``Core.lua``. You can simply access them globally in the Options module. I can do this myself, if you like, but it's probably easier to add it to the PR for you and it'll be more cohesive, too.

Options_Defaults.lua
SacredDuckwhale: True and those will be eradicated as soon as I have more time. If it's not too much of a bother, could you please change it regardless? Thanks :)

Modules/Options/Options.lua
jelveby: No problem, I'll push these changes tomorrow.

jelveby: Ok, I'm having some issues with using `Rarity.CONSTANTS` in the options module. It just throws `table index is nil` when I try to access one of the values from there.

So if instead of

```lua
local TIP_ACTIVATION_HOVER = "TIP_ACTIVATION_HOVER"
local TIP_ACTIVATION_CLICK = "TIP_ACTIVATION_CLICK"
```

I did (not final, just example)

```lua
local ACTIVATION_METHOD_HOVER = Rarity.CONSTANTS.ACTIVATION_METHOD_HOVER
local ACTIVATION_METHOD_CLICK = Rarity.CONSTANTS.ACTIVATION_METHOD_CLICK
```

it would throw `table index is nil` where it's used

```lua
tooltipActivation = {
  type = "select",
  name = L["Tooltip activation"],
  desc = L["If \"On click\" is selected, activating the tracker is done via CTRL + SHIFT + Click, otherwise it's activated with a simple click."],
  values = { -- table index is nil
	  [ACTIVATION_METHOD_HOVER] = L["On hover"], -- table index is nil
	  [ACTIVATION_METHOD_CLICK] = L["On click"], -- table index is nil
  }, -- table index is nil
  get = function() return self.db.profile.tooltipActivation end,
  set = function(info, val)
	  self.db.profile.tooltipActivation = val
	  self:Update("OPTIONS")
  end,
  order = newOrder(),
}, -- tooltipActivation
```

Any thoughts? Is it an execution order thing or something else? `Rarity.CONSTANTS` seem to be defined at least, as far as I can figure.

Debugging is turning out to be a bit of an issue since I'm both new to LUA and addon dev. :P

SacredDuckwhale: The options module is (optionally) loaded on-demand, by the Rarity addon itself, so execution order isn't a problem here. However, the constants for your setting reside in the ``Rarity.CONSTANTS.TOOLTIP`` table, as defined in ``SharedConstants.lua`` (``C`` is assigned ``Rarity.CONSTANTS`` for convenience).

You can debug table using ``/tinspect <tableName>`` or just search manually. This utility is provided by Blizzard and works better than doing ``/dump <tableName>`` manually, which is another option.

jelveby: Don't you just love it when you're an idiot? :D

SacredDuckwhale (approved):

PR #234, opened 2020-12-18
Add Impressionable Gorger Spawn (Loyal Gorger)

nickbayley:
This is a little different to most mounts that drop from NPC's.

The [Impressionable Gorger Spawn](https://www.wowhead.com/item=180583/impressionable-gorger-spawn#comments) has a chance to drop from the [Worldedge Gorger](https://www.wowhead.com/npc=160821/worldedge-gorger). 

This item starts a quest chain where you have to complete a daily quest 7(?) times. After that you will learn the [Loyal Gorger](https://www.wowhead.com/spell=333027/loyal-gorger).

I used the [Torn Invitation](https://github.com/SacredDuckwhale/Rarity/blob/1809d53b8abbf4bf492a43602d2c662b155c23e6/Options_Defaults.lua#L715) as a template, since that is also an item which doesn't directly give you the mount.

### DIFF
--- a/Locales.lua
+++ b/Locales.lua
@@ -1601,6 +1601,8 @@ L["Contained Essence of Dread"] = true
 L["Eternas the Tormentor"] = true
 L["Tower Deathroach"] = true
 L["Decayspeaker"] = true
+L["Impressionable Gorger Spawn"] = true
+L["Worldedge Gorger"] = true
 
 --[[
 					The rest of this file is auto-generated using the WoWAce localization application.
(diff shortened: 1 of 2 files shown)


SacredDuckwhale:
Thanks! I didn't add this item since I misunderstood how it works (apparently). If it's a rare drop then tracking it makes sense.

nickbayley:
I actually got the drop today when I was testing it :P But that was my 10th or so kill, so definitely a rare drop.

PR #241, opened 2021-01-07
DB: Added missing toy Kevin's Party Supplies

godejord:
Added missing 9.0 toy Kevin's Party Supplies [https://www.wowhead.com/item=184447/kevins-party-supplies](https://www.wowhead.com/item=184447/kevins-party-supplies)
Also just aligned another item properly because it was annoying to look at


SacredDuckwhale:
The reformatting probably shouldn't be in here, but thanks :)

PR #247, opened 2021-02-02
Added mount/item/pet filter

kharenis:
I've added a config menu to toggle Mounts, Battle Pets, and Toys & Items to be displayed by the tooltip.
The rationale was huge performance drops when opening the tooltip when I only really want to display mounts at the moment. Filtering out the unneeded collections makes it significantly quicker for me.

image

Also, sorry about the crud added to the .gitignore, it's from Visual Studio :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 # Embedded libraries that are used for local testing only
 Libs/
 *.CMD
+*.suo
+*.json
+*.sqlite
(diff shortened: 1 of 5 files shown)


SacredDuckwhale:
Thank you for your contribution!

Please be aware I force-pushed (overwrote) your commit in order to resolve a conflict without cluttering the history with an additional merge commit. The old branch remains backed up [here](https://github.com/SacredDuckwhale/Rarity/commits/pr-247) in case something went wrong.

This means your local version of the this feature branch will be incompatible with your own remote version (that I overwrote) and shouldn't be built upon if you intend to add something else in the future :)

kharenis:
No problem :) Thanks for the heads up!

Issue #221, opened 2020-11-23
Add fishing nodes for Shadowlands zones to the addon database


SacredDuckwhale:
From Allara's checklist:

- [ ] Fishing nodes (GatherMate2/Constants.lua)

Do we even need these anymore? I think they might've been used for the Sea Turtle. If not, maybe those from 8.3 and other BFA/Legion content patches might be missing?

SacredDuckwhale:
Is there even any item fished from pools? Until I learn of any such thing, there's really no point in adding the fishing nodes to the DB, so I'll save myself the work.

SacredDuckwhale:
Allara:

> For a long time, Sea Turtle was added in new expansions, so we needed new fishing nodes for that. They finally stopped doing that, but we still need to add nodes in order to support user-added items. Users get the option to specify Fishing, a Zone, and “Requires Pool.” Requires Pool internally needs you to add all the nodes, otherwise it won’t detect the pools in the new expansion. (It utilizes tooltip scanning.) It can be difficult to remember, but keep in mind that one of Rarity’s strengths is it’s user-added items. And to fully support that stack of features, you have to keep things going like new fishing nodes, LBZ, etc.

PR #264, opened 2021-02-22
Added missing fishing nodes

godejord:
Added the missing fishing nodes from Shadowlands aswell as some missing from Battle for Azeroth. This should be enough to close issue #221. Not sure if the nodes for mining and such needs to be added as well(?)


SacredDuckwhale:
Mining nodes aren't supported normally; there hasn't been a collectible obtained from mining since Cataclysm IIRC. Fishing pools were originally required for the Sea Turtle mount, but that hasn't been carried over to new expansions either.

However, Rarity allows users to add custom items and one of the methods is "fished from a pool", which does require the fishing nodes to be added. Similar functionality doesn't exist for mining/herbalism/etc so that's the only thing we need to worry about, according to Allara :)

PR #265, opened 2021-02-23
DB: Added Stony's Infused Ruby (pet)

godejord:
Added another missing Shadowlands pet. [Stony's Infused Ruby](https://www.wowhead.com/item=183855/stonys-infused-ruby) from the Venthyr broken mirror system. Haven't personally been able to test this, as I don't have a venthyr, but seems pretty straight forward.

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -1340,6 +1340,23 @@ function R:OnEvent(event, ...)
 			end
 		end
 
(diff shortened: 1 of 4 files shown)


SacredDuckwhale:
Thanks! I guess it's better to have something visible in the addon's tooltip and then reported as not working (if broken) than not having it added at all.

PR #266, opened 2021-02-23
DB: Fixed couple mount coords

godejord:
Both the Kor'kron Juggernaut and the Felsteel Annihilator had missing zones in my GUI, and turns out the UiMapID's have changed somewhere in some patch. Updated for the two that I could see was broken. Could potentially be more


SacredDuckwhale:
Thanks! My first guess would be that those might be the IDs used under the old mapID system, as they seem much higher than anything seen under the new one.

Update: I investigated to see what's going on, and turns it it's simply a typo made in https://github.com/SacredDuckwhale/Rarity/commit/416bc26e47a8301f2eac5e8a99eb5b32bf4494d2#.

Good catch, either way!

PR #267, opened 2021-02-23
DB: Added Smoldering Ember Wyrm. This closes #202

### LINKED ISSUE #202: Add tracking for Smoldering Ember Wyrm
Source: [WowAce](https://www.wowace.com/projects/rarity/issues/384)

I doubt it can be bonus rolled, and not sure if the drop rate has changed? Otherwise this should be fairly easy.

godejord:
Added the Legion mount [Smoldering Ember Wyrm](https://www.wowhead.com/item=142552/smoldering-ember-wyrm) from Karazhan. I know the drop rate is clunky and complicated, but given that this mount is soloable, it's pretty much a flat 20% drop chance. This closes #202

### DIFF
--- a/Locales.lua
+++ b/Locales.lua
@@ -1661,6 +1661,7 @@ L["Luminous Webspinner"] = true
 L["Cache of Eyes"] = true
 L["Stony's Infused Ruby"] = true
 L["Forgotten Chest"] = true
+L["Smoldering Ember Wyrm"] = true
 
 --[[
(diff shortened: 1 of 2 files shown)


SacredDuckwhale:
I agree that no one is going to be running this in a group these days. And even if they were, it'd only affect the "luckiness" display slightly, which I consider a minor (cosmetic) problem at most. Most other drop rates are probably far more inaccurate, so it isn't a great metric to rely on the displayed "luckiness" anyway.

PR #284, opened 2021-03-03
Added an Option to disable achievement toasts. closes #76

### LINKED ISSUE #76: Add option to disable achievement toast when obtaining an item (?)
Source: [Reddit](https://www.reddit.com/r/wow/comments/9zrzk1/rarity_addon/)
> One thing that I've been trying to change for a month now is the achievement-thingy that pops up when you get the item. I have succeded to take away the screenshot-feature, but unfurtunately not with the achievement-like banner.

I'm not sure if this is already possible, hence the "needs confirmation" tag. It should be simple enough to add an option for it if need be.

godejord:
I have added an option to turn off the achievement toasts that Rarity generates when you find an item, as per requested in #76.
The default value is set to true, so it takes manual work to disable it. Hope I haven't missed anything, as this is my first tweak in the options.


SacredDuckwhale:
As mentioned on Discord, the default values should be set in ``Options_Defaults.lua`` :)

SacredDuckwhale:
Sorry for the delay! I wanted to test this first before merging, as there's the potential for things to feel off with UX-based changes. Alas, I noticed the following:

* If the setting is OFF, the achievement notification sound is still played whenever the popup would normally be shown
* Similarly, screenshots are still being taken if the respective setting is ON

As a user, I'd expect the sound to be part of the achievement popup and also be disabled when I turn the notifications off. Perhaps it might also work if both notifications could be toggled separately? That way, users could elect to disable the popup but still be notified with the sound effect? Not sure how useful this would be. It's not very difficult to implement, but it may clutter the config UI a little bit more than it already is (though that's a separate issue).

As for the screenshots... technically, taking a screenshot when an item is found doesn't have to be tied to the achievement popup. *However*, the description for that setting states that screenshots will be taken when the achievement popup is triggered. With this setting enabled, they shouldn't be taken following that logic. From a UX perspective, it also doesn't make sense to take a screenshot without the achievement popup, as the whole point of taking the screenshot is to have the popup screenshotted in order to easily see the item that was obtained.

Thoughts? :D

godejord:
> As a user, I'd expect the sound to be part of the achievement popup and also be disabled when I turn the notifications off. Perhaps it might also work if both notifications could be toggled separately? That way, users could elect to disable the popup but still be notified with the sound effect? Not sure how useful this would be. It's not very difficult to implement, but it may clutter the config UI a little bit more than it already is (though that's a separate issue).

I can understand this. I honestly didn't even realize there was a sound effect tied to the achievement popup. I play without sound, so I just didn't notice. But looking at the code, it seems pretty straight forward to just add the sound under the same setting. I don't really see any point having a separate setting for it, and agree it should just follow the setting for the achievement toast itself.

> As for the screenshots... technically, taking a screenshot when an item is found doesn't have to be tied to the achievement popup. _However_, the description for that setting states that screenshots will be taken when the achievement popup is triggered. With this setting enabled, they shouldn't be taken following that logic. From a UX perspective, it also doesn't make sense to take a screenshot without the achievement popup, as the whole point of taking the screenshot is to have the popup screenshotted in order to easily see the item that was obtained.

This one I did consider, and I decided to not touch it but rather let the screenshot be a separate one. Sure the screenshots are not equally fancy, but depending on which announcement-setting the user has, the screenshot could still take a snapshot of any text announcement as well as the chat where one can see the item looted. I figured one would simply just toggle off both achievement and screenshot if that is how the user wants it. In regards to the description, that definitely needs to be fixed in that case.

SacredDuckwhale:
Sounds alright. I'd then simply change the setting to say "... when an item is found".

Agree on the first point, it's probably best to just disable the sound alongside the popup.

godejord:
Updated the tooltip text for the Screenshot option to remove any confusion, as well as linked the achievement sound to the option to deactive the toast.

Locales.lua
```diff
@@ -631,7 +631,7 @@ L["You entered a achievement ID that is already being used by another item."] =
 L["Instance Difficulty"] = true
 L["Determines which instance difficulties this item may be obtained in. Leave everything unchecked if the instance difficulty doesn't matter.\n\nIf you specified a Statistic ID for this item, the Instance Difficulty is probably meaningless, because all modern statistics already incorporate the difficulty.\n\nYou can check multiple items in this list at once."] = true
 L["Take screenshots"] = true
-L["When on, Rarity will take a screenshot when the achievement alert pops up indicating that you obtained an item."] = true
```
SacredDuckwhale, line 634: Won't this break the ``Options.lua`` entry ``takeScreenshot`` unless the text provided there is also changed?

godejord: Yep, you are absolutely right. Having the strings in multiple locations and not as a single key-value pair still trips me up it seems. Updated the missing one now.

SacredDuckwhale:
Perfect, thank you!

PR #294, opened 2021-03-08
Made achievement toasts unclickable.

### LINKED ISSUE #26: Achievement-style popup causes Lua error when clicked
Source: [WowAce](https://www.wowace.com/projects/rarity/issues/285)

> When Rarity pops up the achievement-like panel like in the first image shown in the Images category, and I click it, the achievements panel opens with somewhat garbled content along with an error:

 

```
Message: ...ns\Blizzard_AchievementUI\Blizzard_AchievementUI.lua:2544: bad argument #1 to 'band' (number expected, got nil)
Time: 03/24/18 19:12:13
Count: 2
Stack: ...ns\Blizzard_AchievementUI\Blizzard_AchievementUI.lua:2544: bad argument #1 to 'band' (number expected, got nil)
[C]: ?
[C]: in function `band'
...ns\Blizzard_AchievementUI\Blizzard_AchievementUI.lua:2544: in function `AchievementFrame_SelectAchievement'
Interface\FrameXML\AlertFrameSystems.lua:421: in function <Interface\FrameXML\AlertFrameSystems.lua:401>
```

> Clicking on actual achievement pop-ups work without errors.
I THINK this started to happen when I installed Pawn, but I can't be sure. I have reported this to the author of Pawn, but as Pawn doesn't interact at all with the achievement window, it doesn't seem likely to be a fault in Pawn. But who am I to tell. :)

godejord:
There is really no idea as far as I can tell as to why these should be clickable in the first place. The original Blizzard achievements open the Achievement frame and shows the new addition there, but these bogus toasts have no ties to that frame. As per now they either result in LUA erros or show a broken achievement frame. This should resolve #26.

### DIFF
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -2631,6 +2631,7 @@ local function RarityAchievementAlertFrame_SetUp(frame, itemId, attempts)
 	shieldIcon:SetTexture([[Interface\AchievementFrame\UI-Achievement-Shields-NoPoints]])
 
 	frame.Icon.Texture:SetTexture(itemTexture)
+	frame:EnableMouse(false)	-- Make achievement toast unclickable
 
 	if attempts == nil or attempts <= 0 then
 		attempts = 1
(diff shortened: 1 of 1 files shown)


SacredDuckwhale:
I like it! Much simpler than messing with Blizzard's UI; no idea why I didn't think of this when I was looking into it at the time.

PR #296, opened 2021-03-09
DB: Added Lucy's Lost Collar

### LINKED ISSUE #276: Add tracking for Lucy's Lost Collar
Source: [Curse](https://www.wowace.com/projects/rarity?comment=2604)

https://www.wowhead.com/npc=175390/dirty-glinting-object apparently has a chance to drop a [pet](https://www.wowhead.com/item=184507/lucys-lost-collar), but is only available during a specific world quest.

godejord:
Added tracking for Lucy's Lost Collar. This was a minor headache, because allthough the Dirty Glinting Object is registered as an NPC, the looting is different from any normal NPC. It uses a different opening spell (called "Looting") and it isn't cought by the normal methods. 

I had to add this as a special case. Rarity.isOpening is hardcoded to only accept spells translated to "Opening" and I didn't want to fake name this spell or expand those just to get past the checks. Detection is therefor similar to the one used for Mail Muncher in Horrific Visions. 

This closes #276.

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -141,6 +141,12 @@ function R:OnSpellcastSucceeded(event, unitID, castGUID, spellID)
 		self:Debug("Finished searching mailbox in a Horrific Vision")
 		addAttemptForItem("Mail Muncher", "mounts")
 	end
+
+	-- Detects opening on Dirty Glinting Object which may contain Lucy's Lost Collar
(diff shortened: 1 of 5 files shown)


SacredDuckwhale:
Thanks! That does sound like the right approach. It's not great, but we can't really do much with the APIs that Blizzard provides.

The entire opening/spell detection stuff is... very adventurous. It definitely could use some reworking, but maybe not right now :joy:

PR #305, opened 2021-03-12
Core: Made faction restriction colors more intuitive

### LINKED ISSUE #304: Color the text displayed on faction-restricted items based on the player's current faction
Example: Pterrordax Egg

Since this item has the ``requiresHorde`` flag set, a notice will be displayed (in red) that the item is only available for Horde players.

So far, so good... right? Unfortunately, this text is *always* red, even when the character is actually a member of the Horde. It would be more intuitive if the text was displayed in green on Horde characters, and red on Alliance (or Neutral) characters.

godejord:
Make a quick fix which should close #304. This makes the notification green if you are on the right faction and stays red when you are not. Found these in the main GUI as well as in the Options. Don't think these show up as NPC tooltips?

### DIFF
--- a/Core/Caching.lua
+++ b/Core/Caching.lua
@@ -22,6 +22,18 @@ function Caching:IsHorde()
 	return Rarity.isHorde
 end
 
+function Caching:IsAlliance()
+	if Rarity.isAlliance == nil then
+		local englishFaction, localizedFaction = UnitFactionGroup("player")
+		if englishFaction == "Alliance" then
+			Rarity.isAlliance = true
+		else
+			Rarity.isAlliance = false
+		end
+	end
+	return Rarity.isAlliance
+end
+
 function Caching:IsReady()
 	return not IsInitializing
 end
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -679,7 +679,7 @@ local function showSubTooltip(cell, item)
 		tooltip2AddLine(colorize(L["Can be obtained with a bonus roll"], yellow))
 	end
 	if item.requiresAlliance then
-		tooltip2AddLine(colorize(L["This mount is only obtainable by Alliance players"], not R.Caching:IsHorde() and green or red))
+		tooltip2AddLine(colorize(L["This mount is only obtainable by Alliance players"], R.Caching:IsAlliance() and green or red))
 	end
 	if item.requiresHorde then
 		tooltip2AddLine(colorize(L["This mount is only obtainable by Horde players"], R.Caching:IsHorde() and green or red))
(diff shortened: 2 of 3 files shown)


SacredDuckwhale:
There's a field in the options (``requiresAllianceT`` and ``requiresHordeT``), but I'm not sure if coloring that makes sense. It's apparently also always printed in red, though I'm not sure where exactly it will be displayed without checking.

godejord:
> There's a field in the options (`requiresAllianceT` and `requiresHordeT`), but I'm not sure if coloring that makes sense. It's apparently also always printed in red, though I'm not sure where exactly it will be displayed without checking.

Those texts are displayed above the specific items in the options. I did color those (: [https://imgur.com/a/veJAVsq](https://imgur.com/a/veJAVsq)

nickbayley:
I know it probably doesn't matter that much, but **Neutral** characters will see green text for mounts marked as `requireAlliance`.

godejord:
Lol, nice catch. Neutral characters is something I forgot existed. Are there any scenarios where neutral characters will have any use for Rarity? Or rather, are they able to wander about in the open world at all? If not, seems a bit redundant to make a whole new function to check for alliance characters just for this

SacredDuckwhale:
That's a good question. I'm not sure what characters are considered neutral, or if the API even differentiates between them. As far as I know, there are

* Pandaren, Death Knights, Demon Hunters, and Allied Races in their starting zone (who may be neutral?)
* NPCs, who might be Neutral (doesn't matter since only players can use Rarity)

In fact, since there's a new starting zone now I actually don't know how things work with the above races anymore.

I suppose you could copy/paste the ``IsHorde`` code and adopt it for Alliance. The entire caching thing needs refactoring, but that isn't something to tackle in this issue :)

Having``IsAlliance`` might not change the logic in practice, but it's more explicit and would make it easier to understand. I feel like it's better to write ``if player is alliance then display this in green, else display it in red`` than ``if player is not horde display this in red, else display it in green`` but maybe that's just me. Basically, if the requirement is ``be of this faction`` and you aren't, then it should always be red. 

It can also be shortened to ``if not IsPlayerAlliance() then red else green`` or similar if verbosity is a concern.

nickbayley:
I can verify that a Pandaren which starts on their original island, rather than the new Shadowlands route, 
is marked as `Neutral` by `UnitFactionGroup("player")`.

I tend to agree about having a separate `IsAlliance` purely to make it easier to read and understand.

godejord:
Added a function to check if the player is alliance. I agree it looks cleaner, but thought it would be unnecessary clutter initially since the functionality was already there ~ish. Either way, tested this on a neutral character as well, and it turns up red for them, as it should.

PR #308, opened 2021-03-15
Strip the colour code from tooltip strings

nickbayley:
This is not usually a problem, as most nodes are not colour coded (although I am not sure if any addons do modify this?).

However, with the addition of the Dirty Glinting Object, this produces an NPC tooltip. For some users, this will be colour coded and therefore it will not properly detect the attempt on Lucy's Lost Collar.

The colour coding will now be stripped if it does exist.

----------------

I'm not happy with where I placed the `StripColourCode` function? Any ideas for a more appropriate place?

Sorry if I haven't made myself clear in the commit/pull request message. I'll try to explain it a bit more if so.

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -835,6 +835,13 @@ end
 
 local tooltipLeftText1 = _G["GameTooltipTextLeft1"]
 
+local function stripColorCode(input)
+	local output = input or ""
+	output = gsub(output, "|c%x%x%x%x%x%x%x%x", "")
+	output = gsub(output, "|r", "")
+	return output
+end
+
 function R:OnCursorUpdate(event)
 	if Rarity.foundTarget then
 		return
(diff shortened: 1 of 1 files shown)


nickbayley:
image

image

The top image doesn't include any colour coding. So the string that is passed is: `A Faintly Glowing Seed`.

The bottom image does, and so the string returned is: `|cffe5b200Dirty Glinting Object|r`.

I'm not sure what addon adds colour to my NPC tooltips (I think it might be ElvUI), but I'm sure others will run into this as well.

Core/EventHandlers.lua
```diff
@@ -835,14 +835,21 @@ end
 
 local tooltipLeftText1 = _G["GameTooltipTextLeft1"]
 
+function StripColourCode(input)
```
SacredDuckwhale, line 838: This pollutes the global namespace, so it'd be better to make it ``local`` or put it in another table somewhere. Since it isn't used anywhere else and we haven't any text-processing modules, I don't know a fitting place to move it to. I guess it'll just have to be refactored out later if it is to be reused.

Also, please use the US spelling instead of the UK version :) While I'm European myself, we should likely be consistent and use the same spelling that's already present in the code, by virtue of the original author being American.

SacredDuckwhale (changes requested):

Core/EventHandlers.lua
nickbayley: Thanks for the input. Changes made.

SacredDuckwhale (approved):

PR #309, opened 2021-03-22
Forgotten Chest has a namesake in Stormsong Valley that should be ignored

nickbayley:
Fixes the issue reported here: https://www.wowace.com/projects/rarity/issues/469

I also removed some duplicated locales and combined the events for the Forgotten Chest to check for both items at the same time.

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -1326,7 +1326,9 @@ function R:OnEvent(event, ...)
 		end
 
 		-- Handle opening Forgotten Chest (Shadowlands, Revendreth chest for Stony's Infused Ruby pet and Silessa's Battle Harness mount)
-		if Rarity.isFishing and Rarity.isOpening and Rarity.lastNode and (Rarity.lastNode == L["Forgotten Chest"]) then
+		if Rarity.isFishing and Rarity.isOpening and Rarity.lastNode and (Rarity.lastNode == L["Forgotten Chest"])
+			and GetBestMapForUnit("player") ~= CONSTANTS.UIMAPIDS.STORMSONG_VALLEY -- Chest with the same name in Stormsong Valley
+		then
 			local names = {"Stony's Infused Ruby", "Silessa's Battle Harness"}
 			Rarity:Debug("Detected Opening on " .. L["Forgotten Chest"] .. " (method = SPECIAL)")
 			for _, name in pairs(names) do
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -1325,12 +1325,12 @@ function R:OnEvent(event, ...)
 			end
 		end
 
-		-- Handle opening Forgotten Chest (Venthyr only chest for Silessa's Battle Harness in Revendreth, Shadowlands)
+		-- Handle opening Forgotten Chest (Shadowlands, Revendreth chest for Stony's Infused Ruby pet and Silessa's Battle Harness mount)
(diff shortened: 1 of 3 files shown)


SacredDuckwhale:
Thanks! I didn't have time to look into the issue much, but if these chests only appear in Revendreth we could probably check for that zone specifically instead? I'll add a code review in a sec to this effect.

Core/EventHandlers.lua
```diff
@@ -1325,12 +1325,14 @@ function R:OnEvent(event, ...)
 			end
 		end
 
-		-- Handle opening Forgotten Chest (Venthyr only chest for Silessa's Battle Harness in Revendreth, Shadowlands)
-		if Rarity.isFishing and Rarity.isOpening and Rarity.lastNode and (Rarity.lastNode == L["Forgotten Chest"]) then
-			local names = {"Silessa's Battle Harness"}
+		-- Handle opening Forgotten Chest (Shadowlands, Revendreth chest for Stony's Infused Ruby pet and Silessa's Battle Harness mount)
+		if Rarity.isFishing and Rarity.isOpening and Rarity.lastNode and (Rarity.lastNode == L["Forgotten Chest"])
+			and GetBestMapForUnit("player") ~= CONSTANTS.UIMAPIDS.STORMSONG_VALLEY -- Chest with the same name in Stormsong Valley
```
SacredDuckwhale, line 1330: Can't we do ``== CONSTANTS.UIMAPIDS.REVENDRETH`` here?

Core/EventHandlers.lua
```diff
@@ -1325,12 +1325,14 @@ function R:OnEvent(event, ...)
...
+		then
+			local names = {"Stony's Infused Ruby", "Silessa's Battle Harness"}
 			Rarity:Debug("Detected Opening on " .. L["Forgotten Chest"] .. " (method = SPECIAL)")
 			for _, name in pairs(names) do
-				local v = self.db.profile.groups.items[name] or self.db.profile.groups.mounts[name]
+				local v = self.db.profile.groups.pets[name] or self.db.profile.groups.mounts[name]
```
SacredDuckwhale, line 1335: I see you replaced ``items`` with ``pets``. I assume this was an unrelated correction and that the code was wrong before, since the item should always have been in the ``pets`` category?

Core/EventHandlers.lua
nickbayley: I haven't unlocked the 3rd tier of the travel network yet which is what is needed to start looting these chests. I don't know if `GetBestMapForUnit("player")` will return only the Revendreth map ID or if there are some others, like the maps in Sinfall etc.
That is why I opted to check for Stormsong.

Core/EventHandlers.lua
nickbayley: It seems this is a symptom of copy and pasting the logic from other containers/chests. I think it isn't necessary on a bunch of them. 

I replaced `items` for `pets` in this one, as I combined the pet with the mount in one if statement, rather than have the duplicated logic we had before.

SacredDuckwhale: Oh, gotcha. Yeah, this whole segment needs streamlining, but that's for another time :)

Core/EventHandlers.lua
SacredDuckwhale: Good point. I also can't check since I'm Night Fae and haven't done any Shadowlands stuff, so I guess we can leave it at this.

Issue #203, opened 2020-10-16
Add holiday-style reminder when Dunegorger Kraulok is available


SacredDuckwhale:
Source: [WowAce](https://www.wowace.com/projects/rarity/issues/425)

It's probably easiest (if a bit inaccurate) to check whether the world quest is currently active. The holiday reminder system uses the calendar API AFAIK, so it may need to be extended somehow.

godejord:
Closed in #311.

PR #311, opened 2021-03-24
Added world quest "holiday" reminders

godejord:
In regards to the issue #203, I have added an option to check if any given world quest is currently active. This is then integrated into the existing holiday reminders, as a way of notifying players when certain items are obtainable. By flagging any relevant items with 'worldQuestId', the reminder should function just like the holiday ones, just with a different display text. If the linked world quest is not active, the item is marked as "Unavailable", and thus can also be hidden from the GUI if the option to hide unavailable items is checked. The option to enable/disable reminders on a specific item is also added on these items with 'worldQuestId'.

Currently I have only implemented the feature on one item; the Slightly Damp Pile of Fur from Dunegorger Kraulok (World boss in Vol'dun). Luckily enough this guy is live this week, so that opens up a window for proper testing without fake ids. The idea is that this can further be implemented on other items such as those tied to Assaults in Uldum and Vale of Eternal Blossom, and possibly others.

I'm uploading this as a draft, as I would appreciate a second glance. Or just feedback in general.

Couple things I haven't gotten around to yet:
- Are the world quest id's the same for Alliance and Horde?
- Need for multiple WQ Id's on a singular item? Ref. above
- What if player hasn't unlocked world quests in that general area?
- - Can items still even drop?
- - Does it matter for the API call?
- Does the API still return the WQ as active after it has been completed? Could affect the Assaults

### DIFF
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -71,6 +71,7 @@ local GetSubZoneText = GetSubZoneText
 local GetContainerItemID = GetContainerItemID
 local RequestRaidInfo = RequestRaidInfo
 local RequestLFDPlayerLockInfo = RequestLFDPlayerLockInfo
+local IsWorldQuestActive = C_TaskQuest.IsActive
 
 -- Addon APIs
 local FormatTime = Rarity.Utils.PrettyPrint.FormatTime
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -1440,8 +1440,10 @@ local function addGroup(group, requiresGroup)
 									numHolidayReminders = numHolidayReminders + 1
 									if numHolidayReminders <= 2 then
 										local text
-										if IsWorldQuestActive(v.worldQuestId) then
-											text = format(L["A world event is currently available for %s! Go get it!"], itemLink or itemName or v.name)
+										if v.worldQuestId then
+											if IsWorldQuestActive(v.worldQuestId) then
+												text = format(L["A world event is currently available for %s! Go get it!"], itemLink or itemName or v.name)
+											end
 										else
 											text = format(L["A holiday event is available today for %s! Go get it!"], itemLink or itemName or v.name)
 										end
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -1292,6 +1292,12 @@ local function addGroup(group, requiresGroup)
 									status = colorize(L["Undefeated"], green)
 								end
 							end
+							-- If item is linked to a World Quest, flag as unavailable if WQ isn't up.
+							if v.worldQuestId then
+								if IsWorldQuestActive(v.worldQuestId) == false then
+									status = colorize(L["Unavailable"], gray)
+								end
+							end
 						elseif v.questId and v.holidayTexture then
 							if Rarity.holiday_textures[v.holidayTexture] == nil then
 								status = colorize(L["Unavailable"], gray)
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -1433,14 +1433,18 @@ local function addGroup(group, requiresGroup)
 								-- Holiday reminder
 								if
 									Rarity.db.profile.holidayReminder and Rarity.allRemindersDone == nil and v.holidayReminder ~= false and
-										v.cat == HOLIDAY and
+										(v.cat == HOLIDAY or v.worldQuestId) and
 										status == colorize(L["Undefeated"], green)
 								 then
 									Rarity.anyReminderDone = true
 									numHolidayReminders = numHolidayReminders + 1
 									if numHolidayReminders <= 2 then
-										local text =
-											format(L["A holiday event is available today for %s! Go get it!"], itemLink or itemName or v.name)
+										local text
+										if IsWorldQuestActive(v.worldQuestId) then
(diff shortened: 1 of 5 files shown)


SacredDuckwhale:
Thanks! Here's my thoughts on the bullet points:

> Are the world quest id's the same for Alliance and Horde?

According to wowhead, yes.

> Need for multiple WQ Id's on a singular item? Ref. above

Probably not. If that ever changes, checking for a table instead of a number value, similar to the ``questId`` field, would be a trivial change.

> What if player hasn't unlocked world quests in that general area?

Then they won't see any reminders, and also attempts tracking for the item won't work.

> Can items still even drop?

Maybe. But even so, it'd affect only a tiny portion of users. Considering the detection through other means is extremely tricky (see Darkmoon Rabbit), it's probably not worth the effort at this time.

As for the last two points:

In another addon I wrote, I simply used this code to check for available world bosses:

````
--- Returns whether or not a given world quest is currently active (and not completed)
local function WorldQuest(questID)
	if not C_TaskQuest or not C_TaskQuest.IsActive then
		return
	end

	return C_TaskQuest.IsActive(questID)
end
````

This has never caused any issues. So I think Blizzard's API simply returns ``false`` after the quest is complete as it's no longer available. ``C_TaskQuest.IsActive(questID)`` doesn't actually seem to query the server whether a given world quest is up on the server, but rather if it's available *to the player*. I've just tested this ingame and it appears to hold true still.

Assaults might also have some other API? Not sure if the world quest is sufficient there, but that's something to worry about later I guess.

**Comment by SacredDuckwhale** [comment]
One other thing I noticed: The holiday reminder appears to show up in Vol'dun only? If this doesn't just happen for me it would mean the usefulness of the reminder is significantly reduced.

**Comment by godejord** [comment]
> > Are the world quest id's the same for Alliance and Horde?
> 
> According to wowhead, yes.

For Dunegorger Kraulok the WQ is definitely the same on both factions. Would have to be verified for other potential candidates for this type of reminder though.

> > Need for multiple WQ Id's on a singular item? Ref. above
> 
> Probably not. If that ever changes, checking for a table instead of a number value, similar to the `questId` field, would be a trivial change.

Looking up the Assaults, there seem to be multiple entries on Wowhead for the same type of Assault. Haven't really studied it, but for me it looked like the intro questline had an identical quest to the WQ, which can be ignored. Leaving this as is for now.

> ... `C_TaskQuest.IsActive(questID)` doesn't actually seem to query the server whether a given world quest is up on the server, but rather if it's available _to the player_. ...

Well, that kinda blows. Isn't an issue in the case of Dunegorger Kraulok though.

> Assaults might also have some other API? Not sure if the world quest is sufficient there, but that's something to worry about later I guess.

Aye, something to look forward to in the next episode.


> One other thing I noticed: The holiday reminder appears to show up in Vol'dun only? If this doesn't just happen for me it would mean the usefulness of the reminder is significantly reduced.

That's weird. I haven't been to Vol'dun a single time while testing this. Can't really say I have touched anything related to zone either. Sure you did a 'reload'? It only notifies you once. :sweat_smile:

**Comment by SacredDuckwhale** [comment]
> That's weird. I haven't been to Vol'dun a single time while testing this. Can't really say I have touched anything related to zone either. Sure you did a 'reload'? It only notifies you once. sweat_smile

I did some more testing, and it doesn't appear to be limited to the current zone. Rather what happens is that the holiday reminder is displayed when I first open the tooltip, instead of on login.

**Comment by godejord** [comment]
> I did some more testing, and it doesn't appear to be limited to the current zone. Rather what happens is that the holiday reminder is displayed when I first open the tooltip, instead of on login.

Is this supposed to be displayed on login? I haven't touched anything related to when these should be triggered. It's just added on top of the existing holiday reminder.

Looking at the original code, current **master**, simply changing one line in GUI.Lua (1394)
from `status = colorize(L["Unavailable"], gray)`
to `status = colorize(L["Undefeated"], green)`
will make some normal holiday items like Big Love Rocket trigger it's holiday reminder. For me that does not show on login, but first when I open the tooltip

**Comment by SacredDuckwhale** [comment]
Yeah, same here... and that's peculiar. I would expect the reminder to be displayed after logging in. Perhaps it would be too messy /easy to miss and so it was decided to only display it when the tooltip is manually opened?

**Comment by godejord** [comment]
> Yeah, same here... and that's peculiar. I would expect the reminder to be displayed after logging in. Perhaps it would be too messy /easy to miss and so it was decided to only display it when the tooltip is manually opened?
> 
> Edit: Either way it's not related to this PR.

I did some investigation, and it looks like it actually _is_ intended to be displayed on login. Unrelated issue, but still something to look into I suppose.

Ref. this bit in Core.lua
```
-- First-time initialization finished
if not Rarity.Caching:IsReady() then
	Rarity.Caching:SetReadyState(false)
	-- Trigger holiday reminders
	self:ScheduleTimer(
		function()
			Rarity:ShowTooltip(true)
		end,
		5
	)
end
```
From what I can tell, the initialization never gets finished(?), and thus the tigger here is never executed.
Debugging this shows `Initialization failed to retrieve 17 item(s)` multiple times before it terminates itself with `Maximum number of cache prime attempts reached`. It never passes `if not Rarity.Caching:IsReady()`..

SacredDuckwhale:
I moved that problem to a separate issue, so as to not derail this PR further :)

It's probably best to simply ignore it here and look into it later.

PR #313, opened 2021-03-25
DB: Removed references to ingame objects to resovle Initialization errors

godejord:
Debugging another issue got me looking into why I was getting 17 Initialization errors. Turns out many of the new Shadowlands items had references to ingame objects via the 'items' field in the DB. I am the one who added most of these, if not all - so seems I made this mess to start with. 

AFAIK this field is useless unless method=use, so a misunderstanding from my part. If someone is able to verfiy that, perhaps this could be added in the DB Verification routine.


SacredDuckwhale:
I suppose I should've seen this problem in the original PR, but I completely missed it... "Shame! Shame! Shame!"

At least I can confirm the field is meant to define the containers that the item can be looted from. It's not a very descriptive name, of course, but you probably already noticed that.

However, I shudder at the thought of anyone looking at the DB verification. That part is really more a proof-of-concept and should never have been committed in its current state :fearful:

PR #320, opened 2021-03-31
Refactor Database Maintenance Helper (and related /rarity verify slash command routine)

### LINKED ISSUE #169: Refactor Database Maintenance Helper (prototype) to implement a streamlined Validation API
* Required fields need to exist, and be named consistently (Id vs ID vs id, etc.)
* Types should be consistent, as well. Right now they're really not, which has caused multiple bugs
* Some fields may be obsolete or could be done away with? (Personal Loot related, I suppose)
* All items should have zones assigned so the tooltip doesn't just show an empty entry
* Waypoints themselves are optional, but should also use the UiMapIDs as constants, and not have hardcoded values like they do now
* Some zones may be missing subzones, which could be especially problematic for the fishing detection. Is there an automatic way of finding out which? May be able to use HereBeDragons migration API from back when Blizzard switched to UiMapIDs
* The existing code is a mess, as it was merely a quick prototype
* Add a button to the UI where it can be run manually?
* Defeat detection? Not sure if this is easy to validate...
* Output in a scrolling frame, instead of in chat? There are so many items that it can't even be read
* Maybe display a summary, as well? Something like X failed, Y succeeded, Z warnings?
* It should be possible to check custom items to avoid errors like the one caused by adding empty item data as a "custom" item (separate issue)

Duckwhale:
May god have mercy on the ``DatabaseMaintenanceHelper``'s soul. This (finally) resolves #169!

Progress:

- [x] Move anything related to DB consistency to the schema/Validation API
- [x] Only show summary and errors to end users (use debug messages for details)
- [x] Streamline database schema definition (it's a mess)
- [x] Rename ``/rarity verify`` slash command to ``/rarity validate`` (do you even English?)
- [x] Update project description to reflect this change in syntax
- [x] Cleanup


Duckwhale:
I think it's better than before... but there's always room for improvement. Any thoughts? @godejord

godejord (approved):
This is great. It seems to be running very smoothly, and at first glance looks easy to maintain and expand. R.I.P. old DBH.
Good work!

Only thing that comes to mind is when I run `/rarity validate` as a user without debug on, I don't really get feedback on _why_ any items failed, just which item. Now, it's very possible that the users that are capable of running the validation also is capable of turning on debug to find the errors, but I fear that this may cause bug reports for a lot of items - and those reports are useless unless we know why they fail as well. Just my two cents 👍

Duckwhale:
I figured it might be too much spam, but on second thought it would be useful to display this. The error messages aren't localized, and not really phrased in a way that would make sense to translate, but I guess that's not a big deal.

godejord:
Ideally only custom items would throw any errors, as items provided by Rarity default should be clean. And if user input-fields eventually get verification too, then errors should be even more rare. With that in mind, localizing etc isn't really necessary

Duckwhale:
Good point either way. I'll just update all the messages to no longer require debug mode. It can always be changed back later if need be.

Issue #252, opened 2021-02-03
Integrate a Lua testing framework


Duckwhale:
It's way overdue; ideally we want to run tests inside the client to make sure the API interactions are working as expected.

Not sure if LuaUnit or similar frameworks will run in the modified Lua environment, but if all else fails I could adapt the UnitTester addon I wrote for another project. Offline testing might also be an option, though mocking the WOW API seems like a lot of unnecessary work when we can just access it directly.

Once this is done, refactoring the legacy codebase can be tackled. This is mandatory if we want to fix long-standing issues such as the LibQTip/UI loading time disaster and related "script ran too long" errors. Just refactoring away at the Core without having tests to rely on has a good probability of breaking absolutely everything, so that doesn't sound very appealing.

Duckwhale:
Which one would work best? There's [busted](https://github.com/Olivine-Labs/busted) for a BDD approach, and [LuaUnit](https://github.com/bluebird75/luaunit) for a more classical, TDD approach.

The former seems to have more community participation, but uses the awkward syntax that's commonly used in the JavaScript world, too.

Duckwhale:
I've looked into busted first, and here's my thoughts so far:

* It's made available *only* through ``luarocks``
* ``luarocks`` is practically impossible to get working on Windows, and IME it just doesn't work unless you want to expend a truly obscene amount of effort on something that should be simple and painless (hello ``npm``)
* Therefore, requiring this to be used for testing would increase the barrier of entry and cause a lot of misery for anyone who tried and isn't a Linux user - which in the WOW world likely describes "most people"
* Requiring nothing but a simple lua interpreter (``lua5.1`` or ``luajit``) seems far lower of a demand
* For the CI pipelines it really doesn't matter, but developers should be able to run tests locally on Windows

Using ``busted`` would be a liability instead of an asset due to its heavy reliance on ``luarocks``. No dice.

Duckwhale:
Got LuaUnit working in just a minute, though I've used it before (years ago). I'm not sure if it can be embedded into the WOW client's Lua environment, but I don't think so since it uses functions like ``require()`` that we'd have to mock.

Duckwhale:
Alright, I was able to mock a bunch of functions that we can't use, like ``math.randomseed`` or ``os.clock``, but whether that's enough remains to be seen. Update: It's not, there are obscure errors inside the LuaUnit code.

I just discovered https://github.com/Jaliborc/WoWUnit, so maybe we can use that for integration test instead.

PR #329, opened 2021-04-02
Split the GUI module into multiple parts to make it easier to handle

Duckwhale:
Let's... get... dangerous! :eyes:  

Roadmap:

- [x] Split the existing code into different files (one per area of responsibility)
- [x] Move library code (LibSink, LibQTip, ...) to separate wrappers in order to streamline the API


Duckwhale:
Since this PR has been a WIP for ages, maybe it's time to split it up to get the changes merged (especially helpful to tackle #350 properly). 

Although I would consider it "unfinished", here's my thoughts on the changes thus far:

* Splitting the GUI into multiple parts based on the individual widgets seems like a good starting point for further refactoring
* The ``Core/GUI.lua`` is still a dumpster fire, but there's too much unrelated functionality that I can't easily move elsewhere
* In order to refactor the non-GUI code that's strewn all over the place, I'd want proper tests (i.e., #252) to avoid breaking everything
* This would require much more work and should likely be a separate GUI-related issue (or several)
* I also like the extraction of string and table utilities, despite it not being part of the GUI per se
* The 3rd party interfaces I don't like, at all
* I tried to gave them domain-specific names, instead of just doing ``TradeSkillMaster.lua`` and ``Rarity.TradeSkillMaster`` or similar
* This doesn't seem like the best design yet as it feels kind of awkward, but I don't want to block this PR just because of it

Any ideas, feedback etc. on the state/progress so far @godejord? Anything important missing that absolutely needs to be in?

We could merge this in despite its "unfinished" state and revisit the remaining problems after we have automated tests (i.e., hopefully soon).

godejord:
This PR is massive, so it's hard to really give any detailed feedback. It does look a lot more manageble and clear at first glance. Unless you have taken some functionality out or broken anything in the process, then I'd say merge it as is, and take it from there. It's likely easier to see which parts are more pressing to refactor in the next iteration once we get a bit used to this new layout.

Also, like you said, this kinda have to get merged before any other issues regarding the GUI gets tackled. No way I want to experience merge conflicts with all this 😆

Duckwhale:
Oh, yeah. It's a lot of changes, but if you read the commit history you'll see I didn't change any GUI code really. I split them into different files, did some minor refactoring, and tried to refactor more but realized it's too convoluted and risky without proper tests.

To proceed with anything else, I'd either have to "finish" this or we just split it off and I do it later, separately. Seems like a less risky approach :joy:

Duckwhale:
Also, I guess if I did break something then I'l have to fix it. But I didn't notice any issues with the UI while testing, and haven't touched anything else.

Duckwhale:
Just to make extra sure I'll push this alongside the recently-added 9.1 toys in an alpha release. Nothing like using players as QA, I guess :laughing:

Issue #327, opened 2021-04-02
Review item database and add missing waypoint data


Duckwhale:
From Discord:
> And force some optional fields like coords.

Since the main window won't display any zone for items without waypoint data, it might be time to make this field mandatory. It doesn't need to have coordinates per se, but a ``mapID`` should exist to make sure a zone name is displayed in the UI.

It doesn't need to be perfectly accurate, as long as it's evident to the player where approximately they can get the item, in case they happen to be in the zone (and they may have hidden items "not in the current zone").

Needs to wait for #320 to be merged first.

godejord:
May have to be creative with the Island Expedition items

Duckwhale:
How so? Doesn't it suffice to simply add the expedition maps (and Boralus/the Horde capital if we care about the crates)?

godejord:
Well, I don't have all the facts on Island Expeditions, but AFAIK some rewards are only tied to certain maps (not even thinking about the weekly creature types). For example the [Island Thunderscale](https://www.wowhead.com/item=166467/island-thunderscale) is believed to only drop from Jorundall. Ref [here](https://www.wowhead.com/island-expeditions-collection-rewards#commonrewards-thunderscale-dragons), and on the comment section of the item itself.

Currently I believe the tracking is just tracking all items on all Island Expedition maps, and updating every individual entry would be a major headache - and is bound to be incomplete or full of errors as the data just isn't there.
I guess adding the zones in a similar manner would be the most doable solution.. Add all maps for all items tied to Island Expedition. 

Then my initial thought was having zones display something like "Island Expeditions" instead of i.e. "Whispering Reef + 10", but I suppose that's just creating more work for no good reason.

Duckwhale:
How about we take the lazy way out and add *only* Boralus/Dazar'alor?

Technically, this is where the expeditions are started from, and also where the crates can be purchased.

godejord:
Sounds like a good compromise! 😅

Issue #130, opened 2019-08-25
Make sure all entries have waypoint data (or at least a zone ID) assigned to them


SacredDuckwhale:
Use the verification (DB maintenance helper) for this. The desired result would be that all items display a zone name, even if no waypoint data exists.  If this doesn't work by default, then the waypoint feature would have to be tweaked.

Issue #129, opened 2019-08-25
Zone names are not displayed if no waypoint data exists


Duckwhale:
Source: [WowAce](https://www.wowace.com/projects/rarity?comment=1957)
> Does anyone understand why (f.e. in Battle Pets section) for Brinstone Algan I see the correct zone (Nazjatar) but for Wriggler, Budding Algan and several other battle pets, the zone is empty. When I check the settings for those battle pets, they do not look different from the ones, where the Zone is displayed properly.

godejord:
Has this changed over the past years? There are plenty items without any waypoints/coordinates that still show their respective zones as long as the zone id is set. I.e. all the items from paragon reputation caches etc.

The items the user is referring to doesn't have the zone id set either, but that's somewhat covered in #130. :)

Duckwhale:
I think "waypoint data" refers to the ``coordinates`` entry, regardless of whether it only consists of a ``mapID``.

If there isn't any such field, I would expect no zone to be displayed if true, and otherwise that Rarity attempts to display the zone based by looking up the ``zoneText`` via ``LibBabble`` (IIRC).

Issue #282, opened 2021-03-03
Refactor the GUI module


Duckwhale:
If this is still too big a task, might have to split it even further. The way it is now, I'd definitely consider it unmaintainable.

Duckwhale:
I settled for splitting this and moving out some unrelated functionality. Everything else will require separate issues and ideally be supported by automated tests in order to avoid breaking things.

PR #332, opened 2021-04-03
A ton of missing waypoints, and validation for item zone #327 

godejord:
I wrote a short validation to check that all items have the `coords` field set, and that it contains a mapID.

Found about 130-150 items without a mapId attached to it, and as I am writing this, I have updated about 80 of those entries to now have a mapId. Also added a couple new entries in the constants obviously.

EDIT: Related issues: #129, #327, #130.


godejord:
Update: I have added mapID's for every possible item that was lacking. Some remaining items however have not had a mapId set, because they are honestly not tied to a zone whatsoever.. The following items can be obained anywhere in the world, and I'd be curious to hear how those can be best solved. If the zone validation is to remain in place, then these entries either need to be an exception to that, or they need some kind of entry for "anywhere in Azeroth".

Missing mapID's:
```
**Pet battle rewards**
-Chain Pet Leash
-Porcupette

**Alchemy**
-Wind-Up Frog
-Undulating Blue Sac
-Baby Crawg
-Ball of Tentacles
-Ridgeback Piglet
-Thaumaturgical Piglet
-Transmutant

**Tailoring**
-Imperial Moth
-Imperial Silkworm

**Other**
-Knockoff Blingtron
-Lil' Bling
-Sky-Bo
```

Duckwhale:
Thank you for the tremendous effort! A few thoughts:

* When items don't have any zone that should be displayed, we could assign a "dummy" zone ID, perhaps ``NONE = 0`` or similar
* The UI could then either display nothing, some value to indicate they don't have a special zone, or any arbitrary string really
* This, however, should be handled in a separate issue/PR (and probably needs to wait for https://github.com/WowRarity/Rarity/issues/282)

Also, could you please do a quick ``rebase`` and ``reword`` the commit messages to be consistent? I know there aren't any official guidelines (yet :eyes: ), but they should probably all use a similar format.

If there's anything in the code itself, I'll start a review on it shortly.

DB/Item.lua
```diff
@@ -150,4 +150,17 @@ function Item:IsMount(entry)
 	return entry.type == CONSTANTS.ITEM_TYPES.MOUNT
 end
 
+function Item:HasZone(entry)
+	if entry.coords then
```
Duckwhale, line 154: I think early exit makes it more readable:

`if not ... then return false end`

or just

`if not ... then return end`

If you agree, could you please change that?

DB/Item.lua
```diff
@@ -150,4 +150,17 @@ function Item:IsMount(entry)
...
+		for _, value in pairs(entry.coords) do
```
Duckwhale, line 155: What does the ``value`` represent? Unless it's a generic table function, I'd use a name that accurately describes the type of object we're iterating over. In this case, probably something like ``waypoint`` or ``waypointData``?

DB/Item.lua
```diff
@@ -150,4 +150,17 @@ function Item:IsMount(entry)
...
+			if type(value) == "table" and value.m ~= nil then
```
Duckwhale, line 156: Early exit?

``if type(waypointData) ~= "table" or not waypointData.m then return false end``

In fact, it may be more readable to split both conditions, or use some auxiliary name like ``local isValidEntry = <condition>`` and then ``if not isValidEntry then return end``? Not sure, just a few quick ideas.
The ``m`` is, of course, another awkward name, but that'll have to be changed individually (not part of this PR) :/

DB/Item.lua
```diff
@@ -150,4 +150,17 @@ function Item:IsMount(entry)
...
+				if tonumber(value.m) then
```
Duckwhale, line 157: Should this check be part of the ``HasZone()`` logic? Type validation seems like it would be separate from checking the existence of required fields, though we haven't streamlined the validation enough yet to say for sure...

Duckwhale (changes requested):

DB/Item.lua
godejord: Right. I forgot to take that out. Copy pasted that from an early protoype. Removed type validation in d872cdd488ff00c51da27fd74d72b4a4750ad488

DB/Item.lua
godejord: Sure. I've renamed it in 039729e9df6633a37a838bf692b3abb8955a5b95.

DB/Item.lua
godejord: I didn't think of that, but it does make sense. I've fixed that part in 039729e9df6633a37a838bf692b3abb8955a5b95.

DB/Item.lua
godejord: The function had some unnecessary logic to it, so I rewrote most of it in 2b30bf11dc726546918879a62c0f240b8378f55f. It now has an early exit for any entries that are missing coords entirely, and it validates the item on the first occurrence of a mapID.

godejord:
> * When items don't have any zone that should be displayed, we could assign a "dummy" zone ID, perhaps `NONE = 0` or similar

Done. Added a dummy zone to constants, and I assigned the dummy zone to all 14 items referenced earlier in this PR.

> * The UI could then either display nothing, some value to indicate they don't have a special zone, or any arbitrary string really
> * This, however, should be handled in a separate issue/PR (and probably needs to wait for #282)

Yeah, not touching that at the moment. Perhaps the 'zoneOverride' could work here. Not sure if that property works without any mapId's attached, but something to look into I guess.

> Also, could you please do a quick `rebase` and `reword` the commit messages to be consistent? I know there aren't any official guidelines (yet 👀 ), but they should probably all use a similar format.

Done and done!

DB/Item.lua
```diff
@@ -150,4 +150,19 @@ function Item:IsMount(entry)
...
 
+function Item:HasZone(entry)
+	if not entry.coords then
+		return false
+	end
+
+	-- It has value in 'coords', make sure it contains a mapID.
+	for _, waypointData in pairs(entry.coords) do
+		if waypointData.m then
+			return true
```
Duckwhale, line 161: Won't this return as soon as *one* of the waypoints has a ``mapID`` assigned? If the goal of the validation is to make sure *all* waypoints care complete then this could be changed to

``if not waypointData.m then return false end``

but I'm not sure if that was the intended behaviour.

Duckwhale:
Also one more thing I forgot to consider, if we use ``0`` (or any other placeholder ID) as ``NONE``, will this break Blizzard's APIs in some way? I'd expect they just return nil, but you never know with how creative they can get...

Duckwhale:
As for the commits, there are a few that seem like they were overlooked:

* https://github.com/WowRarity/Rarity/pull/332/commits/9cc2b85fdf0b677880bea53258b828abf6940469: Added X seems more in line with the others, no?
* https://github.com/WowRarity/Rarity/pull/332/commits/a3a8f08f4948063d80a3995e8eb43ae2606831ee: "This is it"? I'd remove that, and if needed always include additional text in the commit body only
* https://github.com/WowRarity/Rarity/pull/332/commits/e876c0125d3a53fe4950b85d6b0eb01339ec505e: Same for the addendum
* https://github.com/WowRarity/Rarity/pull/332/commits/457c76b14b0bc3aee0b78f5a74a07d5c7e80993d and https://github.com/WowRarity/Rarity/pull/332/commits/d872cdd488ff00c51da27fd74d72b4a4750ad488: Imperative?
* https://github.com/WowRarity/Rarity/pull/332/commits/457c76b14b0bc3aee0b78f5a74a07d5c7e80993d: Is this actually a bug? IIRC trailing commas in Lua are valid. Either way I'd probably ``fixup`` that (via ``rebase``)
* https://github.com/WowRarity/Rarity/pull/332/commits/039729e9df6633a37a838bf692b3abb8955a5b95: Passive? Or whatever that tense is called, I'm bad at grammar :stuck_out_tongue:

DB/Item.lua
godejord: The intention is simply to make sure the entry has a mapId set.

godejord:
> Also one more thing I forgot to consider, if we use `0` (or any other placeholder ID) as `NONE`, will this break Blizzard's APIs in some way? I'd expect they just return `nil`, but you never know with how creative they can get...

Didn't encounter any issues while testing this. AFAIK the invalid UiMapID is already covered in GUI.Lua
`
-- Helper function (to look up map names more easily)
-- TODO: DRY (not sure where this fits best, move after refactoring the rest and delete any duplicates)
-- Returns the localized map name, or nil if the uiMapID is invalid
local function GetMapNameByID(uiMapID)
	local UiMapDetails = GetMapInfo(uiMapID)
	return UiMapDetails and UiMapDetails.name or nil
end
`

godejord:
> As for the commits, there are a few that seem like they were overlooked:
> 
> * [9cc2b85](https://github.com/WowRarity/Rarity/commit/9cc2b85fdf0b677880bea53258b828abf6940469): Added X seems more in line with the others, no?
> * [a3a8f08](https://github.com/WowRarity/Rarity/commit/a3a8f08f4948063d80a3995e8eb43ae2606831ee): "This is it"? I'd remove that, and if needed always include additional text in the commit body only
> * [e876c01](https://github.com/WowRarity/Rarity/commit/e876c0125d3a53fe4950b85d6b0eb01339ec505e): Same for the addendum
> * [457c76b](https://github.com/WowRarity/Rarity/commit/457c76b14b0bc3aee0b78f5a74a07d5c7e80993d) and [d872cdd](https://github.com/WowRarity/Rarity/commit/d872cdd488ff00c51da27fd74d72b4a4750ad488): Imperative?
> * [457c76b](https://github.com/WowRarity/Rarity/commit/457c76b14b0bc3aee0b78f5a74a07d5c7e80993d): Is this actually a bug? IIRC trailing commas in Lua are valid. Either way I'd probably `fixup` that (via `rebase`)
> * [039729e](https://github.com/WowRarity/Rarity/commit/039729e9df6633a37a838bf692b3abb8955a5b95): Passive? Or whatever that tense is called, I'm bad at grammar 😛

Yikes, grammar police 👀 JK. I thought you just ment cleanup the DB tags initially. Clearly I've been sloppy with the commit messages

Duckwhale:
Haha, my apologies. I could probably find more things to improve, but I'll stop looking now :joy:  As they say, perfect is the enemy of good.

Thank you!

Duckwhale (approved):

Issue #326, opened 2021-04-02
Add validation criteria for all fields that have some interaction with each other


Duckwhale:
Some examples:

* Covenant-restricted items require both a flag and a convent ID (used to query Blizzard's API only when actually needed)
* Items obtained from lootable containers expect an ``items`` property for the container(s) you can open to get it
* When waypoint coordinates are set, they should have at least a ``mapID``. When they have x, y values both must exist, etc

There may be others. 

The database schema contains all known fields and if an existing property is unknown, the schema validation will fail.
We can use that as a starting point and simply check each field. in the code (how it is used).

Where there are interactions, they should be checked in the validation logic, by extending the existing API.

Goal: If any of those fields is modified, any field that would also be affected needs to also be modified (or deleted) accordingly to guarantee the addon behaves as expected.

Needs to wait for #320 to be merged first.

Duckwhale:
Am I mistaken or is this now resolved?

@godejord Your PR should cover it all, right?

godejord:
I think I covered all the fields that interact with each other, so this one should be resolved yea 🥳

PR #333, opened 2021-04-08
Validation for field interactions

godejord:
I wrote a series of validations for several of the fields that are dependant on each other. More could be missing, but these were the ones I could think of.

This PR includes
- COLLECTION items need 'collectedItemId'
- ARCHAEOLOGY items need 'raceId'
- USE items need 'items'
- FISHING items need 'zones'
- ZONE items need 'zones'
- NPC items need 'npcs'
- BOSS items need 'npcs'
- 'requiresCovenant' and 'requiredCovenantID' goes together
- Validation to see that X coord is set if Y is set, and vice versa.

Notes:
- Does not include any validation for defeat detection at the moment. That needs to be cleaned up a bit first. Or at least needs a clear line forward. (Replace lockBossName/lockDungeonId with lockoutDetails?) ..Then it can be implemented here to start correcting the faulty, legacy or missing ones.
- Some of these are stolen from Serialization.lua. Serialization validates items before import/export. This section should probably be removed and rather refer to the Schema for import/exports.

Related issue: #326


DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
 	return false
 end
 
+function Item:IsCollectionItem(entry)
+	return entry.method == CONSTANTS.DETECTION_METHODS.COLLECTION
+end
+
+function Item:IsUsingCollectionProperties(entry)
```
Duckwhale, line 172: Is this type of naming an accurate description? IIRC in the first version I named those methods ``IsValidXYZ`` because the function would return whether the sum of the item's properties amount to a valid item of the given type. ``IsUsingXyzProperties`` sounds like there's some sort of provider for Xyz properties?

Not sure what to make of this, perhaps we can find a better description for these types of validation methods? I thought of them as similar to unit tests, though not explicitly labeled as such, but I figure either way we should keep names consistent and agree on a convention here :)

Duckwhale, line 172: Is this type of naming an accurate description? IIRC in the first version I named those methods ``IsValidXYZ`` because the function would return whether the sum of the item's properties amount to a valid item of the given type. ``IsUsingXyzProperties`` sounds like there's some sort of provider for Xyz properties?

Not sure what to make of this, perhaps we can find a better description for these types of validation methods? I thought of them as similar to unit tests, though not explicitly labeled as such, but I figure either way we should keep names consistent and agree on a convention here :)

DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
...
+	return entry.collectedItemId ~= nil
+end
+
+function Item:IsArchaeologyItem(entry)
+	return entry.method == CONSTANTS.DETECTION_METHODS.ARCH
+end
+
+function Item:IsUsingArchaeologyProperties(entry)
+	return entry.raceId ~= nil
```
Duckwhale, line 181: Perhaps this could be narrowed down to actually enforce the field's content to be one of the recognized constants?

DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
...
+function Item:HasUseMethodProperties(entry)
+	return entry.items ~= nil
+end
+
+function Item:IsCovenantSpecificItem(entry)
+	return entry.requiresCovenant ~= nil
+end
+
+function Item:IsUsingCovenantProperties(entry)
+	return entry.requiredCovenantID ~= nil
```
Duckwhale, line 197: Perhaps this could be narrowed down to actually enforce the field's content to be one of the recognized constants?

DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
...
+
+function Item:IsUsingNPCProperties(entry)
+	return entry.npcs ~= nil
+end
+
+function Item:HasSpecialMethod(entry)
+	return entry.method == CONSTANTS.DETECTION_METHODS.SPECIAL
+end
+
+function Item:IsUsingCoordsProperties(entry)
```
Duckwhale, line 232: ``CoordsProperties`` is technically accurate, but a better name might be ``HasWaypoints`` or ``HasWaypointsData``?

DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
...
+	return entry.coords ~= nil
+end
+
+function Item:HasBothXYCoordinates(entry)
+	for _, waypoint in pairs(entry.coords) do
+		if type(waypoint) == "table" then
```
Duckwhale, line 238: I think I would move this to a separate function ``IsValidWaypoint(waypointData)`` and then re-use this in here, so that the waypoints can be validate separately and it's more reusable as a whole. Or perhaps I'm misunderstanding what you intended to validate here?

DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
...
+
+function Item:HasSpecialMethod(entry)
+	return entry.method == CONSTANTS.DETECTION_METHODS.SPECIAL
+end
+
+function Item:IsUsingCoordsProperties(entry)
+	return entry.coords ~= nil
+end
+
+function Item:HasBothXYCoordinates(entry)
```
Duckwhale, line 236: Is this a descriptive name? If you intend to validate *all* the waypoints I'd use ``HasOnlyValidWaypoints()`` or something. If you want to see if it has *any* waypoints, maybe ``HasWaypoints()`` , or maybe you actually meant ``Has[AtLeastOne]ValidWaypoint()``? It's unwieldy, I guess.

Since you only check for x and y coordinates, it appears to be none of the above? Waypoints without a ``mapID`` are incomplete, as ``TomTom``'s API requires ``x``, ``y`` and ``mapID``, while Rarity itself can use the ``mapID`` without coordinates but not the other way around.

DB/Item.lua
```diff
@@ -165,4 +165,83 @@ function Item:HasZone(entry)
...
+	for _, waypoint in pairs(entry.coords) do
+		if type(waypoint) == "table" then
+			if (waypoint.y and not waypoint.x) or (waypoint.x and not waypoint.y) then
```
Duckwhale, line 239: If this is moved to a separate function, you could also make it an early exit again :)

Duckwhale:
What's with the ``DB API`` prefix? Is it ``DB`` or ``API``? APIs would be the modules inside the ``Core``, like ``Serialization``, while DB is the database itself (and also currently includes the schema and validation criteria, which probably isn't ideal and should be moved elsewhere later).

DB/Item.lua
godejord: Well, you did name the first one `Item:IsUsingHolidayProperties(entry)`, so I simply copied the format you used on that one.

I overlooked the `Item:IsValidHolidayItem(entry)` function that's already implemented for HolidayProperties, so I took a slightly different approach. If I am to follow the structure used on holiday items, then there should be 3 functions
Item:IsXYZItem(entry) - Just to check the entry is specific category/method/something
Item:IsValidXYZItem(entry) - Check that all interactions are set, and if needed, have the right values
Item:IsUsingXYZProperties(entry) - Is using fields tied to this case. No other entries should use these properties

If you think this looks right, then I'll simply add the 'IsValidXYZItem' onces and update the schema accordingly

Duckwhale: Obvious disclaimer: My names aren't usually great :joy: 

If I remember correctly, holiday items were a bit of a weird case. I agree with haxing ``IsXyzItem()`` as a shorthand for the basic category check, and ``IsValidXyz()`` can take care of the validation.

Not sure if the ``IsUsingXyzProperties()`` is really needed though? Maybe we can just remove it, as it would be part of the validation logic to check that a field is of the appropriate type, which implies it can't be ``nil`` (since ``typeof(nil) == "nil``.

This means there would only be two functions/checks for each item type, which makes things easier. I'm a big fan of making things easier, but perhaps there's a better way?

godejord: Sure, I don't really mind taking away the `IsUsingXyzProperties()` ones. It's function is basically to warn if someone stuffed things like 'zones' on BOSS items, or 'raceid' on FISHING items, but it's just a warning at this point either way.

Duckwhale: You don't have to take them out, obviously :) Having more validation isn't a bad thing, and if things get too unwieldy we can just move them to a different file/restructure it.

godejord: Rewrote most of the validations, so they generally only use Item:IsXYZItem(entry) and Item:IsValidXYZItem(entry)

godejord: Rewrote most of the validations, so they generally only use Item:IsXYZItem(entry) and Item:IsValidXYZItem(entry)

DB/Item.lua
godejord: Checking for actual valid raceIds in a8bdc6a7bd3929696c5a8daefb117dcae653ad45

DB/Item.lua
godejord: Checking for actual valid covenantIds in e1e77bdd39b5bbc878999175be753afbe2f9af44

DB/Item.lua
godejord: Mostly rewritten in 4b8d97888ef3b226ee87f97c1d9797a2d5590871

DB/Item.lua
godejord: Mostly rewritten in 4b8d97888ef3b226ee87f97c1d9797a2d5590871

DB/Item.lua
godejord: Mostly rewritten in 4b8d97888ef3b226ee87f97c1d9797a2d5590871

godejord:
> What's with the `DB API` prefix? Is it `DB` or `API`? APIs would be the modules inside the `Core`, like `Serialization`, while `DB` is the database itself (and also currently includes the schema and validation criteria, which probably isn't ideal and should be moved elsewhere later).

Yeah good question. I have no idea what that was about. I've done an overhaul of the whole PR and fixed the initial commit messages aswell.

DB/Item.lua
```diff
@@ -150,19 +150,118 @@ function Item:IsMount(entry)
...
+	return true
+end
+
+function IsValidWaypoint(waypointData)
+	if type(waypointData) == "table" then
+		if not waypointData.m then
+			Rarity:Print("Missing mapID")
+			return false
+		elseif (waypointData.y and not waypointData.x) or (waypointData.x and not waypointData.y) then
+			Rarity:Print("X/Y-Coordinates are inconsistent")
```
Duckwhale, line 260: Same here; "inconsistent" doesn't really mean much so I'd probably rephrase this to something like "Waypoint data has incomplete map coordinates". Could also use "invalid" or "partial" here, but that doesn't quite sound as clear (to me, anyway). What do you think?

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:HasUseMethod(entry) and not Item:IsValidUseMethodItem(entry) then
+		Rarity:Print("Found item with use method, but it's items are not valid.")
+		return false
+	end
+
+	if Item:IsCovenantSpecificItem(entry) and not Item:IsValidCovenantItem(entry) then
+		Rarity:Print("Covenant settings are inconsistent")
```
Duckwhale, line 47: See above ("inconsistent" too vague?)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:IsZoneItem(entry) and not Item:IsValidZoneItem(entry) then
+		Rarity:Print("Zone settings are inconsistent")
```
Duckwhale, line 52: See above ("inconsistent" too vague?)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:IsFishingItem(entry) and not Item:IsValidFishingItem(entry) then
+		Rarity:Print("Fishing settings are inconsistent")
```
Duckwhale, line 57: See above ("inconsistent" too vague?)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:HasNPCMethod(entry) and not Item:IsValidNPCItem(entry) then
+		Rarity:Print("NPC settings are inconsistent")
```
Duckwhale, line 62: See above ("inconsistent" too vague?)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:HasBossMethod(entry) and not Item:IsValidNPCItem(entry) then
+		Rarity:Print("BOSS settings are inconsistent")
```
Duckwhale, line 67: See above ("inconsistent" too vague?)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:HasWaypointData(entry) and not Item:HasValidWaypoints(entry) then
+		Rarity:Print("Waypoint data is inconsistent")
```
Duckwhale, line 72: See above ("inconsistent" too vague?)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
 
-	if not Item:HasZone(entry) then
-		Rarity:Print("Zone is missing")
+	if Item:IsCollectionItem(entry) and not Item:IsValidCollectionItem(entry) then
+		Rarity:Print("Collection settings are inconsistent")
+		return false
+	end
+
+	if Item:IsArchaeologyItem(entry) and not Item:IsValidArchaeologyItem(entry) then
+		Rarity:Print("Archaeology settings are inconsistent")
```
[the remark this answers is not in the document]
godejord: This particular one gives additional feedback to the users within `Item:IsValidArchaeologyItem(entry)`.
Will print either "Found archaeology item without raceID property" or "Found archaeology item with invalid raceID" as well as the "Archaeology settings are inconsistent"

DB/Schema.lua
godejord: Gives additional feedback from `Item:HasValidWaypoints(entry)`

DB/Item.lua
```diff
@@ -150,19 +150,118 @@ function Item:IsMount(entry)
...
+		Rarity:Print("Found archaeology item without raceId property")
+		return false
+	else
+		for _, constantRaceIDs in pairs(CONSTANTS.ARCHAEOLOGY_RACES) do
+			if constantRaceIDs == entry.raceId then
+				return true
+			end
+		end
+
+		Rarity:Print("Found archaeology item with faulty raceId")
```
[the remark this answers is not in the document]
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Schema.lua
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Schema.lua
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Schema.lua
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Schema.lua
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Item.lua
```diff
@@ -150,19 +150,118 @@ function Item:IsMount(entry)
...
+			return false
+		end
+	end
+	return true
+end
+
+function IsValidWaypoint(waypointData)
+	if type(waypointData) == "table" then
+		if not waypointData.m then
+			Rarity:Print("Missing mapID")
```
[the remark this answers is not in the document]
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Item.lua
godejord: Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
 		return false
 	end
 
-	if not Item:HasZone(entry) then
-		Rarity:Print("Zone is missing")
+	if Item:IsCollectionItem(entry) and not Item:IsValidCollectionItem(entry) then
+		Rarity:Print("Collection settings are inconsistent")
```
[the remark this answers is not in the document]
godejord: You are right. Rephrased this in 7bb53d56902e48e69116e48dbe8cca91c4fe8179.

DB/Item.lua
```diff
@@ -150,19 +150,118 @@ function Item:IsMount(entry)
...
+
+function Item:IsArchaeologyItem(entry)
+	return entry.method == CONSTANTS.DETECTION_METHODS.ARCH
+end
+
+function Item:IsValidArchaeologyItem(entry)
+	if not entry.raceId then
+		Rarity:Print("Found archaeology item without raceId property")
+		return false
+	else
```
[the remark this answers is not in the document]
Duckwhale: Fair enough :)

DB/Schema.lua
```diff
@@ -28,8 +28,53 @@ function DatabaseSchema:IsValidItem(entry)
...
+		return false
+	end
+
+	if Item:IsArchaeologyItem(entry) and not Item:IsValidArchaeologyItem(entry) then
+		Rarity:Print("Archaeology settings are inconsistent")
+		return false
+	end
+
+	if Item:HasUseMethod(entry) and not Item:IsValidUseMethodItem(entry) then
+		Rarity:Print("Found item with use method, but it's items are not valid.")
```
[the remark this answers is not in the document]
Duckwhale: I know, right? Unfortunately we can't change it without potentially breaking custom items users have added using this field, but it's definitely on the "Most Wanted" list :100: 

Nice attempt, it's certainly very clear now. We probably can't avoid awkwardness entirely, with how the property is named :)

DB/Item.lua
```diff
@@ -150,19 +150,118 @@ function Item:IsMount(entry)
...
+
+function Item:IsValidFishingItem(entry)
+	if not Item:IsValidZoneItem(entry) then
+		return false
+	else
+		return entry.requiresPool ~= nil
+	end
+end
+
+function Item:HasNPCMethod(entry)
```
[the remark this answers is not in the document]
godejord: `Item:IsLootable()` isn't exactly accurate enough for this case. You could say items found in chests etc are lootable too, and containers are sorta lootable.
`HasDetectionMethodNPC()` is the only reasonable option I can come up with at the moment, apart from what has already been mentioned. But yea, if no good option comes to mind, just leave it as is for now..

godejord:
> Could you please modify them to just say `DB:`?

Sure.

Duckwhale:
It's a thing of beauty :P Thanks again for your great contribution!

PR #340, opened 2021-05-24
DB: Added defeat detection to Battlecry of Krexus

godejord:
As the title says, I just added defeat detection to [Battlecry of Krexus](https://www.wowhead.com/item=184318/battlecry-of-krexus)


Duckwhale:
Thanks!

PR #342, opened 2021-05-26
DB: Added defeat detection to Sharkbait's Favorite Crackers

godejord:
Added defeat detection to [Sharkbait's Favorite Crackers](https://www.wowhead.com/item=159842/sharkbaits-favorite-crackers) - the mount from Freehold. Also changed it from BOSS to NPC method, as this is no longer available in mythic keystone and such


Duckwhale:
Thank you!

PR #344, opened 2021-05-27
Added defeat detection to Hopecrusher Gargon

godejord:
Here comes more missing detection. This time the Venthyr only rare mount [Hopecrusher Gargon](https://www.wowhead.com/item=180581/hopecrusher-gargon).


Duckwhale:
Neat, thank you :)

PR #347, opened 2021-06-17
DB: Added defeat detection for Fandral's Seed Pouch

godejord:
As per title, I have been grinding [this stupid toy](https://www.wowhead.com/item=122304/fandrals-seed-pouch) and decided to add the defeat detection along the way.

### DIFF
--- a/DB/Toys/WarlordsOfDraenor.lua
+++ b/DB/Toys/WarlordsOfDraenor.lua
@@ -63,6 +63,25 @@ local wodToys = {
 			["WARLOCK"] = true,
 			["WARRIOR"] = true
 		},
+		lockoutDetails = {
+			mode = CONSTANTS.DEFEAT_DETECTION.MODE_AND,
+			{
+				encounterName = "Majordomo Staghelm",
+				instanceDifficulties = {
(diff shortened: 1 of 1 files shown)


Duckwhale:
Thanks! I suppose Timewalking difficulty/availability can be ignored?

godejord:
Honestly I didn't even think about timewalking. Wasn't even aware Firelands was "timewalkable" 😬 Tried looking through the comments now, but couldn't find any information about this dropping or not..

Duckwhale:
I'd assume it drops all the same from Timewalking raids. The difficulty would be evaluating whether a) Timewalking is active for the raid, b) the player is locked to it, and then c) adding this to the lockout detection.

Seems like a lot of hassle and difficult to test, which is why I figured we could just ignore this possibility. I highly doubt people want to farm the Timewalking version for this item, when they can just run the normal raid again another week.

godejord:
You are probably right in that it should drop even on timewalking. Tracking defeat detection on that is indeed a headache. Neat feature, but headache.

I vote for not implementing that part, and just rethink it if it somehow gets requested later on (which I doubt)

Duckwhale:
Sounds good; I doubt anyone would even notice.

Issue #12, opened 2018-01-23
Add lifetime attempts / account-wide statistics to the UI


SacredDuckwhale:
[WowAce Link](https://www.wowace.com/projects/rarity?comment=184)
>  is there a way to show the number of attempts on items i've already acquired?

Some notes:
1. There appears to be no way to display the stats across characters unless an item is tracked
2. If a usable item has been obtained before Rarity tracked it, there is no way to update the counter even if it could theoretically be derived from statistics / the Achievement API
3. The item in question is a mount - not sure if this applies to all item types?
4. Rather than brute-forcing (and overriding) the actual attempts counter, calculating overall statistics based on a character or account should be done separately
5. This data could then be displayed in the UI, and possibly used to calculate luckiness for items that don't have proper data but could be of interest to the player
6. I don't know (at this time) how much of this functionality already exists, so it needs some investigation before adding anything that may end up obsolete

Duckwhale:
More requests for this feature:

> Hey is there anywhere in rarity i can see how many attempts it took me on the mounts i already have that rarity was tracking while getting them? when i look at recently obtained mounts it says 0 attempts
Source: Discord

And another from [Reddit](https://www.reddit.com/r/wow/comments/jy05lw/rarity_addon_how_to_view_already_obtained_mounts/).

Should probably bump this one up in priority, seeing how it's clearly more useful than I had originally anticipated.

Issue #172, opened 2020-04-02
mount kill count after obtained


Voxxel:
Hi!

I use Rarity for years to track attempts. Now I finally received Astral Cloud Serpent from Elegon. A guildie asked how much I had to kill it before the mount dropped. I tried everything to gather kill count for Elegon from Rarity but I failed to do that.

Where can I find that information within the addon?

SacredDuckwhale:
You can simply check your ingame statistics; they're provided by Blizzard in the second tab of the Achievements UI. Rarity uses those to track attempts so it would be the same count.

Voxxel:
That was the first thing I did but the ingame stats gives me only 4 Elegon kills on each difficulties I don't know why. (I was farming this guy since Pandaria on my main character, and rarity already counted 1xx kills. 

But come on, the question was how could I print a kill count for a boss in Rarity that's already counted by the addon. Shouldn't be that hard.

It's a pretty basic task to get a kill count number from an addon that primarily created to collect kill count numbers, isn't it...

SacredDuckwhale:
You can try opening the Options UI and check the number written under "Attempts" for the respective item, although it's really using the same statistics obtained from Blizzard's API.

This would only change the result if you've killed it on multiple characters and didn't count those manually, as Rarity accumulates statistics gathered from all characters you logged in on with the addon enabled.

If you still want to investigate further, you could open the SavedVariables and search for ``attempts`` under ``Reins of the Astral Cloud Serpent`` (it should be the exact same number), or manually count all instances of the respective statistics under ``accountWideStatistics`` (6797, 6798, 7924, 7923 is 10/25 normal and heroic Elegon, as listed on [wowhead](https://www.wowhead.com/search?q=elegon+kills)). This is all the addon does internally.

Voxxel:
On every kill, a number printed out (by the addon) that counts how many I killed a particular boss. 
Why on Earth should it be so damn hard to check the last number it printed? 

Isn't this addon for counting kills primarily? Why should I dig through lua / savedvariable files? 
I just can't believe it. I use an addon for years to collect kill counts and the end I can't get the number of my attempts when the mount years later finally drops...

SacredDuckwhale:
The easiest way to get the number *after the fact* is exactly as I wrote:

[...]

PR #349, opened 2021-06-26
Expose lastAttempts in options screen for obtained items

conundrum89:
Intended as a stopgap solution to #172. The actual issue seems to just be that the Attempts counter resets when an item is obtained, and the attempt count is stored in lastAttempts instead:

`get = function(into)
    return tostring((item.attempts or 0) - (item.lastAttempts or 0))
end`

So I am simply displaying lastAttempts in a description field above the Attempts / Chance boxes, if lastAttempts > 0, and otherwise displaying "Not Obtained" (it might not always be accurate to say an item has not been obtained if lastAttempts is 0, but I can't seem to find any more useful way to determine whether that's the case).


Duckwhale:
Thanks for your contribution! Apologies for the late reply, I forgot after initially seeing the PR :)

I can imagine there'd be some issues with inaccuracies and possibly the user experience/UI. I'll need to check this out locally and test for a bit before I can say more, which hopefully I'll find time to do very soon (TM)!

conundrum89:
No worries at all, I'm pretty new to WoW addons (have only really done small local modifications until now) so definitely open to feedback or pointers on where things are XD

Duckwhale:
Alright, so I take a look and I think this is a great idea! Unfortunately, there's some problems I can see:

1. If a mount has been learned, it still shows "Not obtained" for me
2. The text and formatting might benefit from some visual improvements (but that's hardly an issue)
3. It obviously only shows the last attempts count, not all of them (which would essentially resolve #12)

So what can we do about these? Let's see...

1. This is tricky and might warrant a separate issue (to be resolved later), especially since ``lastAttempts`` is used in multiple parts of the addon currently... including the GUI that's already being refactored (#329 )
2. I think just rephrasing the text to make it clear it's not a "guaranteed display" but rather "this might hopefully be right" could be enough. Not sure how to improve the formatting at the moment as that appears to be related to AceConfig
3. Again a separate issue, we could add an entirely new "statistics" header and also include this part there. The difficulty here is mostly that we can't query historical data properly yet and would therefore have to rely on similar hacks to this one. And of course the configuration UI is likely the wrong place for displaying statistics anyway, although I wouldn't consider that big deal for a temporary measure such as this

conundrum89:
Thanks for taking a look!

1. As far as I can tell, it will display "Not Obtained" for an obtained mount only if Rarity doesn't know about the attempts that got you the mount. For example, this happens for me on the Jaina mount because I got it during 8.3, when Rarity didn't track it, so I have lastAttempts = 0 for that mount - but I guess it could also happen if Rarity lost data, or you didn't have an import string for the mount when you got it, or probably any myriad of other reasons! Maybe instead of "Not Obtained" I could make it say "No Previous Attempts Recorded"?
2. I'm not very familiar with the AceConfig stuff, but perhaps there's a more appropriate content type than "description". Is there some other text used elsewhere in Rarity that looks how you'd like this to look? I can figure out how to replicate that here.
3. We can display the total number of attempts (we think) a user has ever had on an item pretty easily - that's just stored in Attempts. It would be cool to track the number of attempts each time an item was obtained and display them in a list or something, and maybe even then display "average attempts per drop" etc, but that information doesn't seem to exist historically - lastAttempts is overwritten each time the item is obtained, and doesn't seem to be recorded anywhere else, unless there's a hidden database of attempt counts that I missed in my (admittedly cursory) searching.

Perhaps for now I put this information up the top (where Drop: and Location: etc are printed), so that it's not displayed alongside configuration fields, and expand it to say "Total Attempts: " and "Last Obtained In: " _or_ "No Previous Attempts Recorded" depending on the data we have. Thoughts?

Duckwhale:
1. Yup, I was thinking something along the lines would be better, as it's very likely not all attempts are still available. Even when relying on Blizzard's statistics, we can only see the total number of kills and not when the mount was actually obtained :(
2. No idea, honestly, if you want to improve this you'll have to check their documentation [here](https://www.wowace.com/projects/ace3/pages/ace-config-3-0-options-tables). Unfortunately I've almost never touched that part so I can't really help, but if you don't want to deal with it that's also fine and we can just improve it later - your choice! :)
3. As far as I know, we don't gather many more statistics than the time/date of an attempt. We could probably derive more information from that but it won't be available for every item, and definitely require a better UI to display. That's part of the issue I linked, so it doesn't have to be in this PR at all

Sounds like a good idea! Feel free to improve this until you're happy with it, or just let me know and we can merge as it is.

conundrum89:
Updated! It now displays only relevant fields (Current Attempts, plus Total and Last Attempts only if they're different from Current Attempts) at the top of the listing (after source information). It'll never say "Not Obtained" since we just won't show "last attempts" in that case. Happy to merge it like this if you don't see any issues :)

Duckwhale:
Thank you! I'm sure the UX could be improved further, but that can wait until the Options UI has been cleaned up a bit (soon™).

I'll merge this as-is and we can start rolling it out to people with the next alpha :)

PR #353, opened 2021-06-30
Patch 9.1 database update (mounts)

godejord:
I've begun the fun.

I think that is all the mounts that have been added. A couple of these have reported a very high, if not 100% drop rate, but I'm sure we will get notified of this if that's the case.
Obviously missing lots of waypoints and defeat detection at this point


Duckwhale:
Thank you! I'll push this as an alpha shortly, so we can get early feedback on the 100% drop rate mounts.

PR #377, opened 2021-07-06
DB: Added Intact Aquilon Core and removed Foresworn Aquilon

godejord:
Further research points that this mount is obtained from a quest reward rather than dropping from the NPC. Added the quest item to the database instead, similar to the Necrolord and Venthyr one.

Video proof: [https://www.youtube.com/watch?v=7MN3dM5NFlI](https://www.youtube.com/watch?v=7MN3dM5NFlI)

### DIFF
--- a/DB/Mounts/Shadowlands.lua
+++ b/DB/Mounts/Shadowlands.lua
@@ -489,21 +489,6 @@ local shadowlandsMounts = {
 			{m = CONSTANTS.UIMAPIDS.THE_MAW, n = L["Fallen Charger"]}
 		}
 	},
-	-- ["Foresworn Aquilon"] = {	
-	-- 	cat = CONSTANTS.ITEM_CATEGORIES.SHADOWLANDS,
-	-- 	type = CONSTANTS.ITEM_TYPES.MOUNT,
-	-- 	method = CONSTANTS.DETECTION_METHODS.NPC,
-	-- 	name = L["Foresworn Aquilon"],
-	-- 	itemId = 186483,
-	-- 	spellId = 353877,
-	-- 	npcs = {180032},
-	-- 	chance = 100, -- Estimate,
-	-- 	requiresCovenant = true,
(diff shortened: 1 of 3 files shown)


DB/Mounts/Shadowlands.lua
```diff
@@ -489,21 +489,21 @@ local shadowlandsMounts = {
...
+	-- 	itemId = 186483,
+	-- 	spellId = 353877,
+	-- 	npcs = {180032},
+	-- 	chance = 100, -- Estimate,
+	-- 	requiresCovenant = true,
+	-- 	requiredCovenantID = CONSTANTS.COVENANT_IDS.KYRIAN,
+	-- 	coords = {
+	-- 		{m = CONSTANTS.UIMAPIDS.KORTHIA, x = 47.0, y = 35.5, n = L["Wild Worldcracker"]}
+	-- 	}
+	-- },
```
Duckwhale, line 506: Why is this commented out? If the item is added to another category, can't it be removed? :thinking:

godejord: > Why is this commented out? If the item is added to another category, can't it be removed? 🤔

It most likely can be removed. The item hasn't been available until the reset this week, so my source is PTR data. Just figured I'd leave it there just in case Blizzard suddenly have done some changes to it

Duckwhale: Right, but is there any point to having this in the codebase (since it's part of the MR)? If not, it should likely be removed; if could always be retrieved from the git commit history when the need arises.

godejord: Nah, you are right. Tis' gone now.

Duckwhale: Nice! I'll squash the two commits to get rid of the 2nd (unnecessary) one before merging the changes. You can restore the original entry from your local branch if needed.

PR #380, opened 2021-07-09
Add GitHub Actions workflow to perform static analysis on every commit

Duckwhale:
What gets measured, gets managed... or so people say!


Duckwhale:
Well, obviously I'll have to merge this now, regardless of all the failures. Tackling them inside this PR is way out of scope. It hurts, but there's no way around it, is there? :confounded:

PR #382, opened 2021-07-10
Add SHOW_LOOT_TOAST event handling to detect attempts for Wilderling Saddle

Duckwhale:
This is currently untested and therefore experimental.

Since I can only get one attempt per day, it might be difficult to verify it's working as intended if the first try doesn't immediately solve the problem.


Duckwhale:
I've pushed a new ``r710-alpha-2`` that should include the fix, but I need some feedback that it's working properly before I can officially merge this in. I botched the item ID and had to fix it after using up my only available attempt for the day, so I'm not 100% sure its fully working.

@godejord If you have a Night Fae that isn't locked out and no one else has confirmed it's working, can you give it a try?

Duckwhale:
Seems like it's working: 

image

Source: https://discord.com/channels/659739379115491338/686272973820395542/863488813048660029

godejord:
Neat! I did try this on my night fae character right now, and at first it seemed fine. I did the encounter, and as soon as the loot toast window appeared, an attempt was output to the sink and thus announced. I did however go into the settings right after to manually update my counter (as I know how many attempts I had previously) and I noticed something strange here. Because even thought the announcement got triggered, the item was still listed with 0 attempts. Same could be confirmed in the main GUI window; zero recorded attempts, allthough the item had been flagged as the active one and the minimap icon was of this one.

Duckwhale:
I've seen a similar report on Discord, and I've created #388 to investigate this separately. I'd guess it's either spaghetti code doing its thing, or me being too dumb to use it correctly. Or both :man_shrugging:

PR #383, opened 2021-07-10
Fix attempts detection for items using the COLLECTION method

Duckwhale:
While the bug is fixed, we should likely do some refactoring and write a test to make sure it can't happen again. That's still TODO since I didn't have time.


Duckwhale:
The refactoring is very basic, as I didn't want to touch more without proper tests.

PR #396, opened 2021-07-15
Fix luacheck warnings so the CI pipeline doesn't fail by default

Duckwhale:


Duckwhale:
Well, that was much easier than expected. I guess I fixed a lot of errors simply by virtue of the VS Code luacheck plugin alerting me to them while I was working on other things.

PR #398, opened 2021-07-17
Implement better profiling tools so that we can troubleshoot performance issues more easily

Duckwhale:
This is the stuff I built while investigating the sorting/LibQTip issues. Will likely have to split into separate issues/branches and review the changes in detail, so I'll leave it as a draft for now.


Duckwhale:
Not sure what to do with this mess right now, will have to revisit later.

Issue #100, opened 2019-03-19
Sorting by anything but name causes noticeable lag/freezing when opening the tooltip


Duckwhale:
Source: [WowAce](https://www.wowace.com/projects/rarity?comment=321)
> Hey everyone , past few weeks had fps drop/freeze when was mouseover rarity icon or when it's loading when you log in, couldn't fix it unless was completely reinstalling it and losing all data(which sucks) but today by mistake clicked rarity icon and changed "sorted by" and all freezes gone, started testing stuff looks like sorting by anything but name making freezes/fps drops(sort by zone most busted just 2-3 freeze).
posting here because maybe anyone had similar issue and devs might can fix it.

Duckwhale:
Another report of this suggests there might be a problem with the sorting algorithms used: https://www.wowace.com/projects/rarity/issues/480

Since we've since refactored the Core and they're self-contained in the ``Sorting`` module, it should now be easier to benchmark and ideally improve the sorting times, even for large numbers of items (to future proof against Blizzard's collection madness, and to make sure the sorting is fast even on low-end computers).

In fact, we should probably set a fixed limit for X items, and make sure it's always kept, to guard against regressions and using automated tests (might require a separate issue).

Duckwhale:
Bumping this up in priority, as it now seems easy enough to tackle and it might account for a large number of the issues people have with Rarity's loading time (alongside LibQTip being slow, see https://github.com/WowRarity/Rarity/issues/136 and https://github.com/WowRarity/Rarity/issues/120).

Duckwhale:
![image](https://user-images.githubusercontent.com/16489133/126031474-671c6dca-a635-4ad8-9a7c-b647bda50684.png)

Looks like LibQTip is the main culprit of slow tooltip rendering times, while sorting has a negligible impact on my system. It's still unclear why changing the sort mode affects the time negatively, perhaps both issues affect each other somehow. I've disabled sorting to basically no effect.

I'll investigate both concerns and see where that leads.

Duckwhale:
No sorting (replaced with NOOP):

![image](https://user-images.githubusercontent.com/16489133/126032637-d81c1164-2999-4bb1-8abf-583dcede7cbe.png)

---

Sorting by name:

![image](https://user-images.githubusercontent.com/16489133/126032672-d7b28dd0-7179-4516-a446-7ca713dc262c.png)

Sorting by category, then name:

[...]

PR #409, opened 2021-07-18
Use built-in table.sort function for Sorting

AySz88:
One cause of #100 is that the sort_X functions in Sorting.lua currently use [selection sort](https://en.wikipedia.org/wiki/Selection_sort), which is notoriously slow, as it performs O(n^2) comparisons.

Lua offers `table.sort()`, a built-in for sorting arrays that can take custom comparators. This [SO thread](https://stackoverflow.com/questions/18043803/what-algorithm-does-table-sort-use) says that it is a C implementation of quicksort, but is not a stable sort (i.e. it may arbitrarily reorder ties). If this is a problem, the comparator can be extended to break ties explicitly. (There are also different O(n log n) algorithms that are stable, like [mergesort](https://en.wikipedia.org/wiki/Merge_sort), but that might not be necessary short of a UI for arbitrary user-specified sorts.)

Profiling on my machine, my worst sort (by zone, with no hidden items) is 680 ms, judging by the debug "AddGroup...Sort" outputs. There may be additional improvements available by making the comparators themselves faster, especially `compareZone`.

You might want to take this opportunity to rename `Sorting.sort2`, but to keep the diff focused, I haven't.

### DIFF
--- a/Core/GUI/MainWindow.lua
+++ b/Core/GUI/MainWindow.lua
@@ -982,7 +982,8 @@ local function addGroup(group, requiresGroup)
 							-- OR: At least one encounter must be defeated
 							-- AND: All encounters must be defeated
 							-- (before the item will be displayed as defeated)
-							local usesNewDefeatDetection = item.lockoutDetails and type(item.lockoutDetails) == "table" and #item.lockoutDetails > 0
+							local usesNewDefeatDetection =
+								item.lockoutDetails and type(item.lockoutDetails) == "table" and #item.lockoutDetails > 0
 
 							if usesNewDefeatDetection then -- Resolve the defeat detection using the item's parameters
 								isDefeated = false
@@ -1063,7 +1064,8 @@ local function addGroup(group, requiresGroup)
 
 						-- Support for Defeated items with multiple steps of defeat (supports quests only)
 						if
-							status == colorize(L["Defeated"], red) and item.defeatAllQuests and item.questId ~= nil and type(item.questId) == "table"
+							status == colorize(L["Defeated"], red) and item.defeatAllQuests and item.questId ~= nil and
+								type(item.questId) == "table"
 						 then
 							local totalQuests = 0
 							local numCompletedQuests = 0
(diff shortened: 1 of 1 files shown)


Duckwhale:
Great stuff, thank you for the contribution!

I was thinking of using the builtin ``table.sort`` myself, assuming they'd use a C implementation of quicksort. I didn't have time to research if they also support comparators after profiling yesterday, as I was mostly focused on the LibQTip issue, but I'm glad to see it's all fairly standard :)

I don't think it's a problem that ties aren't resolved deterministically. I suppose we could make adjustments if it proves to be an issue, but I'll cross that bridge when I get there.

I'll have to clean up the leftovers from yesterday before merging this and I want to do some more profiling locally, but improving the worst case time to what's essentially the best case currently will be huge for anyone using these sort methods!

Duckwhale:
Merging in all the smaller changes from yesterday created some merge conflicts. I'll resolve them locally in a bit.

Duckwhale:
For the record, always branch from ``master`` as the individual branches are subject to change until they're ready to be merged. Since you used the branch I worked on yesterday and I didn't notice, merging in your change will be a bit of a headache I'm afraid :joy:

AySz88:
Glad it helped! Sorry for rehashing what you'd known.

About the comparators, I did notice that `table.sort` specifies boolean while the current comparators return int or boolean arbitrarily. It doesn't seem to cause any error outright, though it might cause some conversion overhead, or block some optimization by the interpreter.

Duckwhale:
I think I have to force-push to your branch; I hope it doesn't destroy anything of value. If there's an issue please let me know and don't pull in the updated branch so we can restore the original commits, though I hope it won't be necessary.

Duckwhale:
> 
> 
> Glad it helped! Sorry for rehashing what you'd known.
> 
> About the comparators, I did notice that `table.sort` specifies boolean while the current comparators return int or boolean arbitrarily. It doesn't seem to cause any error outright, though it might cause some conversion overhead, or block some optimization by the interpreter.

Will have to look into the comparators, but we can easily update them to return ``false`` or ``nil`` instead of ``0`` if needed. We'd probably need to do some benchmarking first to see if it changes anything, but I'd assume it doesn't really change much since they all evaluate to ``false``.

AySz88:
> I think I have to force-push to your branch; I hope it doesn't destroy anything of value. If there's an issue please let me know and don't pull in the updated branch so we can restore the original commits, though I hope it won't be necessary.

There isn't anything there other than this commit, no. I was expecting the PR to merge into your branch rather than into master? Is that not a function that github supports? Sorry about the complications. ^^;

Duckwhale:
The problem is that the branch you wanted to merge in was just me experimenting and benchmarking; I've since split it into several smaller features that have been merged (except for the base branch). This means you now have the WIP commits and not what you had yesterday, but I'm trying to untangle the mess somehow :P

Duckwhale:
Merged via #411, thanks again!

PR #411, opened 2021-07-18
Use the builtin table.sort (quicksort) instead of the slower selection sort algorithm that's currently used by default

Duckwhale:
Extracted from #409  to avoid the headache with rebasing to the various branches.


Duckwhale:
Sorting by zone text is still very slow, but that's because of the convoluted logic behind it.

PR #438, opened 2022-01-03
Added pickpocket collectibles

### LINKED ISSUE #249: Revisit pickpocketing collectibles that are disabled (but still exist in the codebase)
Allara:
> 3. They’re commented out because either I couldn’t implement them (i.e. the pickpocketing items were too much work) or I decided not to (i.e. the fishing items). I think if you look into the pickpocketing one you’ll find that the list of NPC IDs is absolutely massive, hundreds and hundreds of them, and no user was willing to put that list together. So I left the data I had in case someone decided to do it, but commented it out. If you think the items are useful and can be implemented correctly, feel free to put them in. Or delete them! Probably time to delete them haha.

If there's a way to obtain the NPC IDs from wowhead, adding them should be doable. Otherwise, might as well delete them? No need to keep more cruft around than absolutely functionally necessary.

godejord:
Finally got around to adding the three dice-thingys from pickpocketing. I've added the sources currencly listed on Wowhead, but as mentioned in the issue (#249), there are likely more NPCs than we know off. Still beats no tracking at all, and this should cover many of the more popular grind spots for these toys. Drop rates are just average guesstimates.

This closes #249.


Duckwhale:
Pretty neat, thank you!

Issue #386, opened 2021-07-10
Add GitHub Actions workflow to enforce consistent formatting across all files when a PR is updated


Duckwhale:
Goals:

- [x] Before merging in any changes, we want to make sure the formatting is consistent everywhere (see #331)
- [x] When changes have been made to legacy files, the CI will notify us of the inconsistency, so that it can be fixed separately (if unrelated to the PR)
- [x] If this requires elaborate setup, like installing ``luarocks``, it could be easier to just do it after running ``luacheck`` (for every commit?)

Roadmap:

- [x] Research what CLI commands need to be run
- [x] Update the ``luacheck.yaml`` workflow OR create a separate ``luaformat.yaml`` one (TBD)

Issue #331, opened 2021-04-03
Add LuaFormatter configuration file


Duckwhale:
The default format isn't quite ideal, so being able to control the formatting properly would be useful.

Link: https://github.com/Koihik/LuaFormatter

Duckwhale:
Somehow the formatting is seriously messed up in many files, and it makes editing them difficult as randomly, there will be tons of whitespace changes. Bumping this up as it's a real pain when working on the huge legacy files.

PR #449, opened 2022-01-24
Update the list of NPC IDs for the Infinite Timereaver mount

ithappenedagain:
Updated Infinite Timereaver bosses that were updated for 9.1.5 that had some dungeons rotate out and others in for previous TW events.


Duckwhale:
Apologies for the delay in reviewing this, things have been rather hectic over here.

Anyway, I fixed the formatting so that the check passes :) I also tried to undo the large number of whitespace changes, but was ultimately unsuccessful. Since standardizing that properly is a separate issue (see #331 and #386) I'll have to merge it as-is and fix the formatting later.

Thank you again for your contribution!

PR #450, opened 2022-01-25
Added tracking for Bonehoof Tauralus

### LINKED ISSUE #343: Revisit tracking for Bonehoof Tauralus
The [Bonehoof Tauralus](https://www.wowhead.com/item=182075/bonehoof-tauralus) mount was added then removed from the DB as it was thought to not be implemented on release. Fast forward half a year and this mount is very much available ingame, allthough the conditions for it to drop is a bit complex.

Still a rare drop it would be nice to track if possible, so should look into this..

godejord:
Finally got around to looking at this again, and with the new information presented, I have implemented tracking for this badboy. Getting a valid attempt requires having a quest flagged as completed, as this represents having the companion active.

This closes #343.


Duckwhale:
Good stuff! Thank you very much!

PR #451, opened 2022-01-25
Flagged "other" covenant items as unavailable

### LINKED ISSUE #345: Automatically hide covenant-specific items from the Main Window when they can't be obtained
Source: Discord
> One thing that's been on my mind for a while; should we default hide items tied to covenants when you are not a part of that covenant? Like don't show items that only drop for Necrolords if you are a Venthyr. Kinda like class specific items work

godejord:
As mentioned in #345, items that require players to be of a specific covenant to be obtained are effectively unavailable if the player is currently in favor of another covenant. Added some code to flag these items as such if that is the case.

This closes #345 if merged.


Duckwhale:
Thanks! I tried to amend the commit message, but GitHub's weird PR branch structure seems to prevent me from pushing it to your repo. To work around this, I'll create a new PR with your commit and the updated message, then close this one after merging it. You'll still be credited as the author, since it's the same commit - just updated with my title and commit message :)

PR #457, opened 2022-02-05
Adjustments to Viable Cobra Egg

godejord:
Added defeat detection and removed the group settings on the [Viable Cobra Egg](https://www.wowhead.com/item=160832/viable-cobra-egg)


Duckwhale:
Thank you! As you probably noticed, I added some formatting check that requires running ``lua-formatter`` before each commit 🐙 

While I can easily fix the formatting, perhaps you want to take this opportunity to see if you can apply it yourself? This would allow me to determine whether it's easy enough to use, or whether I need to add some instructions.

The fastest way would be to use a Linux system (or WSL on Windows), install [LuaFormatter](https://github.com/Koihik/LuaFormatter) and then run the ``autoformat.sh`` script.

PR #476, opened 2022-05-01
Fix an incorrect negation in the TSM module

Duckwhale:
That's clearly an oversight/rookie mistake. It must have been many years since I wrote that code, though I remember making this particular mistake before...

Anyway: Seems like luacheck recently added the capability for detecting this type of error, so the CI is now failing (as it should)!. However, my local version was outdated and subsequently didn't raise the issue, which is why it went unnoticed for so long.


Duckwhale:
The formatting workflow fails due to an unrelated issue. Merging regardless, since there's no way to get both fixes in simultaneously without making a mess.

PR #485, opened 2022-08-12
Add GameTooltip hooks for Quest Rewards

teelolws:
Allows Rarity info to be shown on GameTooltips from Quest Givers and the Quest Log. Moved the code that handles the Quest Reward to a reusable function.

Questgiver:
image

Quest Log:
image

At the bottom I left commented out code for Emissaries. Currently the text is injected above the Quest Item so I leave it to you to finish that if you wish to use it:
image

### DIFF
--- a/Core/GUI/GameTooltipHooks.lua
+++ b/Core/GUI/GameTooltipHooks.lua
@@ -278,11 +278,8 @@ _G.GameTooltip:HookScript("OnTooltipSetUnit", function(self)
 	end
 end)
 
--- TOOLTIP: ITEMS IN INVENTORY
-
-hooksecurefunc(GameTooltip, "SetBagItem", function(self, bag, slot)
+local function processItem(id)
 	local blankAdded = false
-	local id = GetContainerItemID(bag, slot)
 	if id then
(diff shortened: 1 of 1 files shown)


Duckwhale:
Thank you for the contribution! Let me just quickly fix the formatting :)

I'm not sure about the emissary tooltip, isn't the item already displayed when hovering over the reward box itself? Either way, we can leave it in and decide later. It's a bit messy, but that's the least of my worries when it comes to this codebase 😅

teelolws:
> I'm not sure about the emissary tooltip, isn't the item already displayed when hovering over the reward box itself?

Only when looking at the quest or the item in my bags. With this change I'll be able to see it on the Callings UI in the Garrison Landing Page, so I can avoid going to the NPC if its a calling reward I don't need.

Duckwhale:
Alright, but I guess if it was to be enabled I would append everything to the end, so that the Rarity addition doesn't disrupt the text flow of "this will reward you with" <the reward> that Blizzard intended (see last screenshot). I don't know how easy this is to do, however.

PR #489, opened 2022-09-26
Update the drop rate of Ruby Droplet

borgiman:
According to https://www.wowhead.com/item=104159/ruby-droplet the drop rate is 3% not 4%.


Duckwhale:
Thanks! Not sure what's with the autoformat error, I'll just try running it again...

Duckwhale:
That failure turned out to be an upstream issue, resolved by #490 - I'll update this PR manually so that it can be merged without errors.

PR #496, opened 2022-10-27
Fix OnCursorUpdate event handling to account for 10.0 changes

Duckwhale:
This event has been removed entirely. Using CURSOR_CHANGED instead appears to be working in a similar fashion, though more testing is needed.


Duckwhale:
I can't guarantee this will work in every scenario, but since there's no way of testing that doesn't involve dogfooding... I'll just merge it and see what happens, I guess 🤷

PR #517, opened 2022-12-13
Added Framework for Dragonflight collectibles

godejord:
The journey has begun. Added the framework (item category, icon, constants, settings, etc..) for Dragonflight.


Duckwhale:
Thank you!

Issue #515, opened 2022-12-09
Add Dragonflight mounts to the database


godejord:
Here is a list of new mounts that should be looked into and potentially added. I have gone over the full list and filtered out irrelevant items.
Note that some of these are still missing accurate drop rates.

- [ ] [Verdant Skitterfly](https://www.wowhead.com/item=192764/verdant-skitterfly) Prerequisite: Renown 25 with Dragonscale Expedition. Source is available before getting the required renown, so needs to take this into account
- [x] [Reins of the Liberated Slyvern](https://www.wowhead.com/item=201440/reins-of-the-liberated-slyvern) (#521)
- [x] [Plainswalker Bearer](https://www.wowhead.com/item=192791/plainswalker-bearer) (#521)

Duckwhale:
Looking into the Skitterfly mount, it seems to require looting [yet another object from the ground](https://www.wowhead.com/object=376587/expedition-scouts-pack).

As for the renown check, I think this might work: 

* [C_MajorFactions.GetMajorFactionIDs](https://wowpedia.fandom.com/wiki/API_C_MajorFactions.GetMajorFactionIDs) to find the right ID
* [C_MajorFactions.GetCurrentRenownLevel](https://wowpedia.fandom.com/wiki/API_C_MajorFactions.GetCurrentRenownLevel) to check the renown level

So far that's completely untested (I don't even have DF), but it should be easy enough to manually verify ingame.

cyriun:
excluding renown, is the verdant skitterfly similar to the spectral mawrat's tail from shadowlands?

Duckwhale:
> excluding renown, is the verdant skitterfly similar to the spectral mawrat's tail from shadowlands?

I can't say for sure, but if it's an object lootable from a container on the ground then yes - it should use the same method.

cyriun:
used the same method and it works for every expedition scout pack looted regardless of renown, just got to add the renown check now.

PR #521, opened 2022-12-21
Added some DF mounts

godejord:
Added 2/3 of the new mounts, as per #515.

This PR also includes the UiMapID's for the four new zones


Duckwhale:
Awesome, thank you!

PR #540, opened 2023-03-27
df pet - 10.0.7

cyriun:
also removed some of the groupsize parameters for shadowlands rares

### DIFF
--- a/DB/Mounts/Shadowlands.lua
+++ b/DB/Mounts/Shadowlands.lua
@@ -45,8 +45,6 @@ local shadowlandsMounts = {
 		spellId = 312753,
 		npcs = { 166679 },
 		chance = 100,
-		groupSize = 5,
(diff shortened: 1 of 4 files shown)


Duckwhale:
Thanks! There are three different changes in this PR, which seem to be only semi-related (if that):

1. A `groupSize` update for Shadowlands rares
2. New DB entry for Ashenwing
3. Achievement update for the Forbidden Reach rares

I'll take care of this by merging them separately in a second, but please submit only one set of (related) changes per PR :)

The purpose of this policy is to maintain a somewhat clear history, so that later anyone can see what each PR did without having to read all the commits that it included, e.g., while investigating a bug or trying to understand the motivation for a given change.

cyriun:
thanks for clarifying, will do that from now on

Issue #513, opened 2022-12-08
Reevaluate group sizes for items under BfA & Shadowlands


godejord:
With the release of DF, some items that previously required a group under the BfA and the Shadowlands category may now be soloable.

PR #547, opened 2023-03-29
Update the groupSize settings of various pets from past expansions

cyriun:


cyriun:
merging this should close issue #513 
could probably re-evaluate the bfa raids again after the next patch, most should be soloable by then

PR #549, opened 2023-03-30
Added Raszageth appearance to DB

cyriun:


Duckwhale:
Uhm... should this be part of the core addon? Seems like something that might be fairly niche, i.e., better as an export string.

cyriun:
what do you mean?

cyriun:
i might be wrong but instead of adding a new raid mount into the game, so far they are adding mount appearances so they can be used with dragon riding i guess, i believe theres another coming in the next raid tier

https://www.wowhead.com/news/new-highland-drake-elementium-customization-from-aberrus-raid-in-patch-10-1-332195?webhook

Duckwhale:
I don't know what this is exactly, but it looks like a skin/appearance for a mount (not a collectible mount itself). There's currently no precedent for anything like it being in the addon database, the closest thing I know is transmog appearance and those aren't included. Usually, if people want to track them they can import the items as an export string (serialized form of the item data).

If there is interest in tracking these, we might want to provide a set of export strings so that people can import them separately?

cyriun:
yeah its pretty much a skin, up to you

Duckwhale:
I'll close this as I don't think it's feasible (nor required) for Rarity to track every single cosmetic item out of the box. Thanks, anyway!

If there's interest in these, someone can create a set of export strings and post it on Discord/Curse, just like we have for illusions.

PR #552, opened 2023-04-01
Add more Dragonflight/10.0.7 pets to the database

cyriun:


DB/Pets/Dragonflight.lua
```diff
@@ -119,6 +119,48 @@ local dragonflightPets = {
...
+		method = CONSTANTS.DETECTION_METHODS.NPC,
+		name = L["Scruffles"],
+		npcs = { 200584 },
+		itemId = 193364,
+		spellId = 375047,
+		chance = 25,
+		creatureId = 189115,
+		groupSize = 5,
+		equalOdds = true,
+		coords = { { m = CONSTANTS.UIMAPIDS.THE_FORBIDDEN_REACH, x = 58.6, y = 48.8, n = L["Vraken the Hunter"] } },
```
Duckwhale, line 134: It looks like some of these phrases are missing in the localization table?

DB/Pets/Dragonflight.lua
```diff
@@ -119,6 +119,48 @@ local dragonflightPets = {
...
+		method = CONSTANTS.DETECTION_METHODS.NPC,
+		name = L["Wakyn"],
+		npcs = { 200960 },
+		itemId = 191930,
+		spellId = 371594,
+		chance = 20,
+		creatureId = 188821,
+		groupSize = 5,
+		equalOdds = true,
+		coords = { { m = CONSTANTS.UIMAPIDS.THE_FORBIDDEN_REACH, x = 45.9, y = 79.7, n = L["Warden Entrix"] } },
```
Duckwhale, line 148: Also missing in L?

cyriun: i believe they should have been added in with a different mount?

cyriun: #539 these dont need to be added in a 2nd time right?

Duckwhale: Oh, right. I didn't realize they were the same rares - never mind then!

PR #553, opened 2023-04-02
Add another 10.0.7 pet

cyriun:


Core/EventHandlers.lua
```diff
@@ -1728,6 +1728,21 @@ function R:OnEvent(event, ...)
 		Rarity.isFishing = false
 		Rarity.isPool = false
 
+		--Handle Disgusting vat fishing
+		if Rarity.relevantSpells[405274] == "Disgusting Vat Fishing" and (Rarity.lastNode == L["Disgusting Vat"]) then
```
Duckwhale, line 1732: What is the purpose of this condition? The first part is always true (spell is part of the database), isn't it?

Looking at the other nodes, it seems `previousSpell` might be used to make sure the spell that was cast is the one we want (here, Disgusting Vat Fishing)... Not sure if that's sufficient in this case, though.

DB/Pets/Dragonflight.lua
```diff
@@ -161,6 +161,21 @@ local dragonflightPets = {
...
+		cat = CONSTANTS.ITEM_CATEGORIES.DRAGONFLIGHT,
+		type = CONSTANTS.ITEM_TYPES.PET,
+		method = CONSTANTS.DETECTION_METHODS.SPECIAL,
+		name = L["Emmah"],
+		zones = { "2151" },
+		spellId = 375511,
+		itemId = 193853,
+		creatureId = 189694,
+		sourceText = L["Fished from Disgusting Vat located within the Zskera Vault"],
+		requiresPool = true,
```
Duckwhale, line 174: If this item requires a pool, shouldn't the detection logic check for `Rarity.isPool`?

Duckwhale (changes requested):
~~Some localization phrases seem to be missing.~~

Core/EventHandlers.lua
cyriun: iirc i tried previous spell first but couldnt get it too work, was probably forgetting to add something, not sure

cyriun:
trying to fix this one but i have no clue whats happening with this thing, cannot get it to track attempts from fishing at the disgusting vat but it works fine from any other regular dragonflight pool

Duckwhale:
Does it make sense to complete #608 first and then take another look at this one?

cyriun:
yes although those pets and toy is pretty much done, just abit of clean up left

Duckwhale:
If I understand correctly, this is not working and the problem is that we need to detect fishing from the "pool", correct?

My approach would be to `/etrace` and enable the debug log to see if the pool detection fails, but I can't find the damn thing 😂

cyriun:
> If I understand correctly, this is not working and the problem is that we need to detect fishing from the "pool", correct?
> 
> My approach would be to `/etrace` and enable the debug log to see if the pool detection fails, but I can't find the damn thing 😂

yea so far it works with any other dragon isles pool but not the disgusting vat pool, i do need to look at this again when i get time. The way i find it easily is using the fishing node tracking on the minimap it should pop up, if its not in that vault i usually swap to an alt.

Duckwhale:
As far as I can tell, this is already working? The only issue is that there's a weekly lockout, but adding that is simple.

image

Here's the updated version (with lockout handling) if you want to take a look:

* https://github.com/WowRarity/Rarity/commits/pr-553-backup

Does this work for you? Keep in mind that fishing detection is a bit wonky (you'll need to hover over the vat before looting).

DB/Pets/Dragonflight.lua
Duckwhale: Answer: It's not needed here, since we don't rely on the builtin fishing method (see Mr. Pinchy).

Core/EventHandlers.lua
Duckwhale: Just to recap: This condition is pointless and will be removed. The problem with the current/previous spell variables was that `UNIT_SPELLCAST_SENT` is required to update them, but fishing from the vat doesn't trigger this event. Presumably, it's a special case where Fishing (the spell) is initiated by the server after interacting with the object itself, and not by the client.

Duckwhale (approved):

Duckwhale:
Sorry for taking so long to get this landed! Finally, Emmah is coming home 🐟🐟🐟 (in the next release, that is)

PR #558, opened 2023-07-12
fixes error caused by 10.1.5 updates

elkewtf:
This needs testing by someone else (works for me) but this should fix the error caused by the latest update from 10.1.5 that removed the artifact history ready function

### DIFF
--- a/Core.lua
+++ b/Core.lua
@@ -197,7 +197,6 @@ do
 		self.db.RegisterCallback(self, "OnProfileReset", "OnProfileChanged")
 		self.db.RegisterCallback(self, "OnProfileDeleted", "OnProfileChanged")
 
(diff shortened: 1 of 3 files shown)


GurliGebis:
Are all of these changes needed?
It looks like it removed more than just the event handler and registration

elkewtf:
> Are all of these changes needed? It looks like it removed more than just the event handler and registration

No it should have only been 3 total lines removed. 1 in each of the files that I edited. I wonder if my version was slightly different. HOld please. I'll reclone and make a new PR.

GurliGebis:
You could just amend the commit and force push the branch 🙂
But that works as well 👍

elkewtf:
Yeah except I have no idea how those others changes were pulled in. Haha I didn't trust it.

PR #559, opened 2023-07-12
Fix the archaeology history bug from 10.1.5

elkewtf:
This should be better

### DIFF
--- a/Core.lua
+++ b/Core.lua
@@ -197,7 +197,7 @@ do
 		self.db.RegisterCallback(self, "OnProfileReset", "OnProfileChanged")
 		self.db.RegisterCallback(self, "OnProfileDeleted", "OnProfileChanged")
 
-		RequestArtifactCompletionHistory() -- Request archaeology info from the server
+		self:ScanAllArch("DoEnable")
 		RequestRaidInfo() -- Request raid lock info from the server
 		RequestLFDPlayerLockInfo() -- Request LFD data from the server; this is used for holiday boss detection
 		C_Calendar.OpenCalendar() -- Request calendar info from the server
--- a/Core/Collections.lua
+++ b/Core/Collections.lua
@@ -303,7 +303,6 @@ end
 -------------------------------------------------------------------------------------
 
 function R:ScanAllArch(event)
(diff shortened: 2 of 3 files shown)


Duckwhale:
Thank you! Looking into the patch changes, I don't see any reason why they removed this. I'm assuming it's no longer needed.

Originally, Rarity would request the archaeology project history and then populate its internal cache on load. If the data is now available immediately, there's no need to request it (good) but the data should still be fetched - which, it appears, is still being done in some other places (e.g., on `CURRENCY_DISPLAY_UPDATE`). The last part actually seems like a hack; presumably there was no suitable event when the code was originally written. Either way, that's unrelated to this PR and should be fixed independently.

Duckwhale:
Actually, the `CURRENCY_DISPLAY_UPDATE` handler only updates archaeology fragments and not projects. So this PR removes the only location [outside of `OnProfileChanged`] where the projects are scanned. This may cause some issues with tracking archaeology items (or not?) - I'm investigating if that's true and whether we can just perform a (delayed) manual update on load.

dratr:
You can just fetch the data on initialization now it looks like - see 10.1.5 changes to Interface/AddOns/Blizzard_ArchaeologyUI/Blizzard_ArchaeologyUI.lua here   https://github.com/Gethe/wow-ui-source/commit/da4222fb0cb363a4c18d0ed58aead54cca182ccc#diff-215dac632eb1478e1faccc193fe7e38ea6ad83f15b6f9af953bfe24a7b3deb1e

Duckwhale:
> You can just fetch the data on initialization now it looks like - see 10.1.5 changes to Interface/AddOns/Blizzard_ArchaeologyUI/Blizzard_ArchaeologyUI.lua here [Gethe/wow-ui-source@da4222f#diff-215dac632eb1478e1faccc193fe7e38ea6ad83f15b6f9af953bfe24a7b3deb1e](https://github.com/Gethe/wow-ui-source/commit/da4222fb0cb363a4c18d0ed58aead54cca182ccc#diff-215dac632eb1478e1faccc193fe7e38ea6ad83f15b6f9af953bfe24a7b3deb1e)

Thanks, that's exactly what I did. It seems to be working, so I'll merge this with the added update on load.

Issue #514, opened 2022-12-09
Add Dragonflight toys to the database


godejord:
Here is a list of new toys that should be looked into and potentially added. I have gone over the full list and filtered out irrelevant items.
Note that some of these are still missing drop rates, so could potentially be guaranteed drops; others have lacking sources at the moment. There just isn't enough data to find online yet

- [x] [Talisman of Sargha](https://www.wowhead.com/item=200857/talisman-of-sargha)
- [x] [Notfar's Favorite Food](https://www.wowhead.com/item=200160/notfars-favorite-food)
- [x] [Personal Shell](https://www.wowhead.com/item=198409/personal-shell)
- [x] [A Collection Of Me](https://www.wowhead.com/item=200148/a-collection-of-me)
- [x] [The Super Shellkhan Gang](https://www.wowhead.com/item=200999/the-super-shellkhan-gang)
- [x] [Primalist Prison](https://www.wowhead.com/item=200198/primalist-prison)
- [x] [Infected Ichor](https://www.wowhead.com/item=200178/infected-ichor)
- [x] [Mage's Chewed Wand](https://www.wowhead.com/item=200249/mages-chewed-wand)
- [ ] [Primal Invocation Quintessence](https://www.wowhead.com/item=200636/primal-invocation-quintessence)
- [ ] [Everlasting Horn of Lavaswimming](https://www.wowhead.com/item=200116/everlasting-horn-of-lavaswimming)
- [ ] [Generous Goblin Grenade](https://www.wowhead.com/item=200142/generous-goblin-grenade)

cyriun:
Added a few of these here #541

Duckwhale:
I just noticed one item that was added in that PR has a 50% drop rate on wowhead: https://www.wowhead.com/item=202042/aquatic-shades

Do we even need to track this? I assume @godejord filtered it out because of the drop rate. It can hardly be considered a "rarity"?

Sorry for not catching this sooner 👀

Duckwhale:
> i believe the personal shell toy is also 50%

Some wowhead comments indicate the rare might be bugged. If Rarity shows repeated attempts then (many) people will believe it and continue killing the rare, wasting their time. Since the drop rate is also very high, it might be best to just remove this item?

cyriun:
yeah just scrap it

Duckwhale:
2 of the 3 remaining items seem to not be relevant after all, going by wowhead data.

[...]

PR #599, opened 2023-08-06
Add Everlasting Horn of Lavaswimming (toy) to the database

cyriun:
adding Everlasting Horn of Lavaswimming toy. Too get a more accurate tracking on this toy you would have to track the event completion but for now its tracked by killing the mob "Inferna the Bound" at the end of the event "Siege on Dragonbane Keep".

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -428,9 +428,9 @@ function R:OnCombat()
 	end
 end
 
--- Handle quest turnins: Only used to detect world quests used for outdoor world bosses. It's not ideal, but probably more reliable than the loot lockout quest (which may or may not already be completed when the UNIT_DIED event is fired)
-local worldBossQuests = {
+local worldEventQuests = {
 	[52196] = "Slightly Damp Pile of Fur", -- Dunegorger Kraulok
+	[70867] = "Everlasting Horn of Lavaswimming", -- Scalebane Keep (scenario completion)
 }
 
 function R:OnQuestTurnedIn(event, questID, experience, money)
@@ -438,7 +438,7 @@ function R:OnQuestTurnedIn(event, questID, experience, money)
 		"OnQuestTurnedIn triggered with ID = " .. questID .. ", experience = " .. experience .. ", money = " .. money
 	)
 
-	local relevantItem = worldBossQuests[questID]
+	local relevantItem = worldEventQuests[questID]
 	if not relevantItem then
 		return
 	end
--- a/DB/Toys/Dragonflight.lua
+++ b/DB/Toys/Dragonflight.lua
@@ -380,6 +380,17 @@ local dragonflightToys = {
 			{ m = CONSTANTS.UIMAPIDS.THE_PRIMALIST_FUTURE },
 		},
 	},
+	["Everlasting Horn of Lavaswimming"] = {
+		cat = CONSTANTS.ITEM_CATEGORIES.DRAGONFLIGHT,
+		type = CONSTANTS.ITEM_TYPES.ITEM,
+		isToy = true,
+		method = CONSTANTS.DETECTION_METHODS.SPECIAL,
+		name = L["Everlasting Horn of Lavaswimming"],
+		itemId = 200116,
+		chance = 40, --guess
+		sourceText = L["This toy is obtained at the end of the Siege on Dragonbane Keep event."],
+		coords = { { m = CONSTANTS.UIMAPIDS.THE_WAKING_SHORES, x = 24.2, y = 70.1 } },
+	},
 }
 
 Rarity.ItemDB.MergeItems(Rarity.ItemDB.toys, dragonflightToys)
--- a/Locales.lua
+++ b/Locales.lua
@@ -1988,6 +1988,8 @@ L["Gooey Snailemental"] = true
 L["Dropped from the final bosses of Froststone Vault Primal Storm."] = true
 L["Combine 50 Leftover Elemental Slime to create the Gooey Snailemental."] = true
 L["Froststone Vault Primal Storm"] = true
+L["Everlasting Horn of Lavaswimming"] = true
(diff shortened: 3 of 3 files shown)


cyriun:
been a few months since i've contributed on here so let me know if im missing anything, thanks

Duckwhale:
Thanks! There's probably a better way to track this that also works if you don't actually loot the boss. The comments on wowhead say that the item can be obtained by just AFKing, which to me indicates that it's probably tied to the scenario itself.

We can still use this and improve the detection later, though, if anyone wants to check the event ingame (using `/etrace`).

cyriun:
yes i would say also tracking the container is better than the boss because then you still get an attempt from afking the event but i had just copy pasted the ids from an import string haha

Duckwhale:
Did you still want to implement the scenario detection I outlined in #514, or should I just push my version?
I really don't want to discourage you from experimenting and learning, so feel free to give it a try (it's just one extra function).

If you want to take a look, here's what I did: [wip-pr-599](https://github.com/WowRarity/Rarity/tree/wip-pr-599) (Ideally, use only for cross-referencing)

cyriun:
yeah i could take a look and ive been trying to but i think the event only happens ever 2 hours or so and i keep missing it due to having other stuff to do.

Duckwhale:
No problem, completely up to you. Just let me know when/if we can merge this (or my version, whichever you prefer).

cyriun:
is line 60 relevant for debugging or the tracking altogether?

Duckwhale:
> is line 60 relevant for debugging or the tracking altogether?

What line is that? If it's `RegisterEvent`, then yes: https://wowpedia.fandom.com/wiki/API_Frame_RegisterEvent

cyriun:
ok yes so i think thats where mine went wrong, i seen "self:RegisterEvent("QUEST_TURNED_IN", "OnQuestTurnedIn")" and used that but i think it should've still worked so im not sure

cyriun:
ok nvm the onquestturnedin function does work, i must have done something slese wrong last time i tried it

cyriun:
> No problem, completely up to you. Just let me know when/if we can merge this (or my version, whichever you prefer).

whichever one you think works better, they probably are pretty much the same though, from my understanding

Core/EventHandlers.lua
```diff
@@ -582,6 +582,22 @@ function R:OnIslandCompleted(event, mapID, winner)
 	end
 end
 
+local Scalebane_Keep_QuestID = 70867
```
Duckwhale, line 585: Since this is a constant, we should probably capitalize it (for consistency's sake)?

``local SCALEBANE_KEEP_SCENARIO_COMPLETION = 70867``

Core/EventHandlers.lua
```diff
@@ -582,6 +582,22 @@ function R:OnIslandCompleted(event, mapID, winner)
...
+function R:OnQuestTurnedIn(_, questID)
```
Duckwhale, line 586: It seems there are now two functions `OnQuestTurnedIn`, with the second one overwriting the first.

The first one is used to detect the BFA world boss that drops [Slightly Damp Pile of Fur](https://www.wowhead.com/item=174842/slightly-damp-pile-of-fur). If you want to use this event (instead of `SCENARIO_COMPLETED`), you will have to modify the existing event handler:

1. Add the `SCALEBANE_KEEP_SCENARIO_COMPLETION` ID to the existing list of relevant quests
2. Probably rename it, since `worldBossQuests` would no longer be an accurate description

Other than that, it should "just work" since the existing handler would select the toy if it's in the list.

Duckwhale (changes requested):
You need to modify the existing `OnQuestTurnedIn` handler if you want to use that event :)

Core/EventHandlers.lua
cyriun: ok thats a bit confusing but i think i understand, i would have to add a different function name in the register event section under "QUEST_TURNED_IN" to not overwrite the existing one?

> 1. Add the `SCALEBANE_KEEP_SCENARIO_COMPLETION` ID to the existing list of relevant quests

But also add something similar to to world boss quests? Don't think i quite understand that part

cyriun: but from what i do understand i would do something like 
`local eventQuests = {
	[70867] = "Everlasting Horn of Lavaswimming",
}`

Duckwhale: > ok thats a bit confusing but i think i understand, i would have to add a different function name in the register event section under "QUEST_TURNED_IN" to not overwrite the existing one?
> 
> > 1. Add the `SCALEBANE_KEEP_SCENARIO_COMPLETION` ID to the existing list of relevant quests
> 
> But also add something similar to to world boss quests? Don't think i quite understand that part

If you re-use the existing event handler, no additional function is needed. Once the toy is added to the list, the event handler should automatically add an attempt for it once the quest was completed. This approach requires minimal changes.

cyriun: ohh ok gotcha, simple then

cyriun: also what would be a suitable rename for "worldBossQuests". Something like "eventQuests" perhaps?

Duckwhale: > also what would be a suitable rename for "worldBossQuests". Something like "eventQuests" perhaps?

Uhm... I guess that's fine, but maybe `worldEventQuests` instead of `worldBossQuests`?

cyriun: please tell me im not forgetting anything in that latest commit haha

Duckwhale: Looks good to me. There isn't much to change with this approach. Assuming the event always fires on scenario completion, this version's even simpler than my original idea which required a new event handler. I consider that a good thing :)

Duckwhale:
Merging with some cosmetic changes (streamlined commit history and removed an outdated comment). Thanks!

PR #606, opened 2023-08-12
Add Holoviewer toys to the database

cyriun:
only added 2/3 of the holo viewer toys because the scarlet queen one is around 60% drop rate according to wowhead.


cyriun:
i have to rebase right?

Duckwhale:
> i have to rebase right?

I already did it, hopefully without breaking anything :S

PR #621, opened 2023-08-21
Add chance to DisplayText also

Rubio9:
[Contributing a patch I've kept locally for a couple years]

Add "chance" to DisplayText, so the same calculated chance (displayed as "Likelihood" in the GUI list) will also display in chat log (or wherever else this is sent).


Core/Announcements.lua
```diff
@@ -18,27 +18,37 @@ function Announcements:AnnounceAttemptForItem(item)
...
+		local chance
+                if item.method == CONSTANTS.DETECTION_METHODS.COLLECTION then
+                        chance = (item.attempts or 0) / (item.chance or 100)
+                        if chance < 0 then chance = 0 end
+                        if chance > 1 then chance = 1 end
+                        chance = chance * 100
+                else
+                        local dropChance = Rarity.Statistics.GetRealDropPercentage(item)
+                        chance = 100 * (1 - math.pow(1 - dropChance, attempts))
+		end
```
Duckwhale, line 30: This seems like something we could easily test, but to do that it should likely be moved to a separate function?

I'll take care of the tests if you could extract it; the function can just live in the `Announcements` module for now :)

Duckwhale (changes requested):
The localization needs to be fixed, and ideally the chance computation extracted (for testability). Otherwise, looks good to me :)

Edit: The formatting is also off, but I'll fix that if you don't want to bother with it.

Core/Announcements.lua
```diff
@@ -18,27 +18,37 @@ function Announcements:AnnounceAttemptForItem(item)
...
 
 		if item.lastAttempts then
 			attempts = attempts - item.lastAttempts
 		end
 
 		if total <= attempts then
 			if attempts == 1 then
-				displayedText = format(L["%s: %d attempt"], itemName or item.name, attempts)
+				displayedText = format("%s: %d attempt - %.2f%%", itemName or item.name, attempts, chance)
```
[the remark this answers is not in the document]
Duckwhale: Actually, I noticed that this format would conflict with the one for repeated tracking. Let's just keep it as you suggested.

Core/Announcements.lua
Duckwhale: I guess we can always do that later, if needed.

Duckwhale (approved):

Issue #500, opened 2022-11-12
Container items are no longer tracked reliably after the latest patch (?)


Duckwhale:
There have been multiple reports of this problem:

> Why Rarity addon stopped printing attempt numbers for [Knockoff Blingtron]? It is still in the list of not found pets.
Source: [Curse](https://www.wowace.com/projects/rarity?comment=2816)

> Since prepatch, it hasn't been counting Strange Humming Crystal from Kirin Tor Chests nor Thistleleaf Adventurer from Dreamweaver Provisions.
Source: [WowAce](https://www.wowace.com/projects/rarity/issues/494)

And more on Discord (I suspect they're all referring to the same problem):

* https://discord.com/channels/788119147740790854/788119800617369611/1040366005249118259

---

Looking into this, it seems something wonky is going on:

* Usually, ``BAG_UPDATE`` triggers (twice) and Rarity should detect the attempt after opening a container
* Sometimes it now seems to not trigger at all, in which case ``SHOW_LOOT_TOAST`` was the only event I saw (may depend on the item?)
* Or they might trigger "too early" (before the container loot has been obtained), if I didn't miss it...
* In such a case, Rarity will fail to pick up the new items since they aren't present when the inventory is scanned
* However, in theory it should pick up the attempt when the bags are scanned again later (but it doesn't?)

I guess this needs more investigation, but it's tricky because you need to find container items to open (which are not ubiquitous) AND you can only test it once per item, but also each container may have different rules (loot toast vs bag update...) so it's difficult to generalize.

Duckwhale:
Another theory I had was that the cursor event changes (#493 ) somehow affected the detection for container items, but I can see no way that these two different detection methods should interact. They're two separate event handlers and the inventory scanning logic should not interact with the "node detection" code that is based on tooltip scanning...

mdavidson37:
Debug output from opening a Cracked Fel spotted Egg:

Rarity: LOOT_READY with target: NO TARGET
Rarity: Locking session for 1 second(s) to prevent duplicate attempts
Rarity: LOOT_READY with target: NO TARGET
Rarity: Session is locked ignorning this LOOT READY event
Rarity: Bag Update
Rarity: Setting last count to nil
Rarity: Setting last count to nil

I still have two of the rays to collect - neither had their attempt counts updated.

mdavidson37:
[...]

PR #622, opened 2023-08-23
Fix no loot tracking after using mailbox

Rubio9:
10.x added new events to handle open/close frames for many game UI frames—things like the guild/bank, the auction house, and the mailbox. The old events still exist; some actions trigger both old and new events, some trigger only the old event, and some only trigger the new event. (Most likely, more events will cut over from old to new in future game releases.)

This adds EventRemapping to basically convert new style events to old style events, allowing existing event handler code to work with either type of event.

Most notably here, the MAIL_CLOSED event no longer fires when leaving the mailbox; this left Rarity thinking the player is at the mailbox, so it would ignore loot events.  After EventRemapping, Rarity sees "MAIL_CLOSED" again, hopefully fixing a longstanding issue where loots would be ignored, seemingly inconsistently.

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -54,6 +54,7 @@ local DebugCache = Rarity.Utils.DebugCache
 
 function EventHandlers:Register()
 	self = Rarity
+	local WOW_INTERFACE_VER = select(4, GetBuildInfo())
 
 	self:UnregisterAllEvents()
 	self:RegisterBucketEvent("BAG_UPDATE", 0.5, "OnBagUpdate")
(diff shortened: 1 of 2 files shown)


Duckwhale:
Thanks! For reference, see #500. Review coming in separately in a bit.

Core/EventHandlers.lua
```diff
@@ -1148,7 +1155,36 @@ function R:OnResearchArtifactComplete(event, _)
...
 
+-- 10.x added events PLAYER_INTERACTION_MANAGER_FRAME_HIDE and PLAYER_INTERACTION_MANAGER_FRAME_SHOW
+-- which currently fire either in addition to, or instead of, older events like MAIL_SHOW or MAIL_CLOSED.
+-- This maps new events onto old ones, so there's just one (old-style) event to check against.
+-- (Note many PlayerInteractionTypes are defined but not all are necessarily used; we add the ones most
+-- likely to be relevant here, whether currently being triggered or not.)
+
+local EventRemapping = {
+	["PLAYER_INTERACTION_MANAGER_FRAME_HIDE"] = {
+		[Enum and Enum.PlayerInteractionType and Enum.PlayerInteractionType.Banker       or ""] = "BANKFRAME_CLOSED",
```
Duckwhale, line 1166: You can fix the luacheck CI errors by adding the global `Enum` to the luacheckrc config file in the root directory :)

Duckwhale (commented):
I suspect we'll have to streamline the handling of different client versions if classic clients are to be supported, but since that's a far-off milestone currently I don't have any objections to your approach here. I probably wouldn't even have bothered with the distinction since `RegisterEvent` will throw an error if the event doesn't exist (i.e., on classic clients), but either way is fine :)

To fix the CI errors, you can update the list of globals in luacheck's config and also run [StyLua](https://github.com/JohnnyMorganz/StyLua)).

Core/EventHandlers.lua
Rubio9: Fixed CI errors.

PR #661, opened 2023-10-11
Implement tracking for Box of Rattling Chains

cyriun:
im going to assume there will be a better way to do this. Probably having the table of layer ids in a different file like how the covenant ids are.


Duckwhale (commented):
Thanks! I've a few suggestions that might be worth considering. (see review comments)

Core.lua
```diff
@@ -673,6 +673,66 @@ function R:IsAttemptAllowed(item)
 		return false
 	end
 
+	local _, _, _, _, _, _, _, _, _, LfgDungeonID = GetInstanceInfo()
+	local torghastLayerIDs = {
```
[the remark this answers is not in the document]
cyriun: are you referring to something like this?
image

or do you want it to be more specific listing the torghast instance and layer = id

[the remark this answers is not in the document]
Duckwhale: I figured it could just be a mapping from "human readable name" to "LFG dungeon ID", same as the others.

Or what did you have in mind? [This DBC listing](https://wow.tools/dbc/?dbc=lfgdungeons&build=10.0.5.47660#page=1&search=torghast) should have all the IDs and names AFAICT.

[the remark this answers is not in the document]
cyriun: my interpretation was formatting it like everything else in sharedconstants but all the layer ids are just under 1 table, instead of having a unique name for each id

DB/Toys/Shadowlands.lua
```diff
@@ -458,6 +458,36 @@ local shadowlandsToys = {
...
+			153011,
+			152995,
+			169859,
+			156015,
+			153174,
+			155250,
+		},
+		itemId = 187075,
+		chance = 50,
+		requiresLayerID = true,
```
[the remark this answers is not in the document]
cyriun: i didnt quite understand this format but i think i have something similar or maybe its correct, we'll see

Core.lua
```diff
@@ -673,6 +673,17 @@ function R:IsAttemptAllowed(item)
 		return false
 	end
 
+	local _, _, _, _, _, _, _, _, _, LfgDungeonID = GetInstanceInfo()
+	print("Torghast Layer " .. LfgDungeonID)
```
Duckwhale, line 677: If this is meant to remain as debug output, it should use `Rarity.Debug` so that it can be hidden.

Otherwise, it'll have to be removed so as to avoid spamming the chat frame :)

Core.lua
```diff
@@ -673,6 +673,17 @@ function R:IsAttemptAllowed(item)
...
+	if item.requiredDungeons[CONSTANTS.LFG_DUNGEON_IDS.TORGHAST_LAYER] then
+		if tContains(CONSTANTS.LFG_DUNGEON_IDS.TORGHAST_LAYER, LfgDungeonID) then
+			print("Toy drop tracking for Torghast Layer " .. LfgDungeonID)
```
Duckwhale, line 680: Same here

Core.lua
```diff
@@ -673,6 +673,17 @@ function R:IsAttemptAllowed(item)
...
+		else
+			print("Not in Torghast Layer.")
```
Duckwhale, line 682: And here

DB/SharedConstants.lua
```diff
@@ -264,6 +264,59 @@ C.FACTION_IDS = {
 	ISKAARA_TUSKARR = 2511,
 }
 
+C.LFG_DUNGEON_IDS = {
+	TORGHAST_LAYER = {
```
Duckwhale, line 268: I was actually thinking there would be one entry per layer, and then you could use a simple lookup. Your solution should also work, although technically `tContains` may be slower as it iterates the whole table.

Up to you, I guess, since it's unlikely to make a difference in this case: The branch will only ever be executed inside Torghast, although we could theoretically try to benchmark to see if it's measurable at all. But ehh... whatever.

I would rename this to `TORGHAST_LAYERS`, though, since you're not mapping each layer but all of them.

Duckwhale (changes requested):
The debug output needs to be fixed (or removed). Everything else is at your own discretion :)

DB/Toys/Shadowlands.lua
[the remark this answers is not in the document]
Duckwhale: To expand on the original format:

```lua
requiredDungeons = {
  [Rarity.Enum.LFG_DUNGEON_IDS.TORGHAST_LAYER_1] = true,
  [Rarity.Enum.LFG_DUNGEON_IDS.TORGHAST_LAYER_2] = true,
  [Rarity.Enum.LFG_DUNGEON_IDS.TORGHAST_LAYER_3] = true,
  -- ... etc
  }
```

where the individual IDs are from the shared constants.

But it's all the same to me, no need to change it if you want to keep the old format. If there's a reason to access the IDs directly (which seems unlikely as Shadowlands is old content), we can easily refactor the lookup table to change its format.

Core.lua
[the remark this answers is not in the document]
Duckwhale: See the other comments; both methods work although having just one table means we can't access the IDs directly.

But it really won't make a difference here, unless Blizzard adds new collectibles to Torghast. That, I think, is rather unlikely.

DB/SharedConstants.lua
Duckwhale: Update: I just noticed that holiday events are also using these IDs, which are currently hardcoded. For example, the Hallow's End or Love is in the Air events. These could be moved to the shared constants as well, which means we would probably have to go with the "one mapping per dungeon ID" line here?

Core.lua
cyriun: ill probably remove that one since its abit spammy but yeah i also forgot to change it from print to debug, thanks

DB/SharedConstants.lua
cyriun: well, theres already some freezing lag in torghast so a little more cant hurt

Core.lua
```diff
@@ -674,12 +674,11 @@ function R:IsAttemptAllowed(item)
 	end
 
 	local _, _, _, _, _, _, _, _, _, LfgDungeonID = GetInstanceInfo()
-	print("Torghast Layer " .. LfgDungeonID)
 	if item.requiredDungeons[CONSTANTS.LFG_DUNGEON_IDS.TORGHAST_LAYER] then
-		if tContains(CONSTANTS.LFG_DUNGEON_IDS.TORGHAST_LAYER, LfgDungeonID) then
-			print("Toy drop tracking for Torghast Layer " .. LfgDungeonID)
+		if tContains(CONSTANTS.LFG_DUNGEON_IDS, LfgDungeonID) then
```
Duckwhale, line 678: Wouldn't this return `true` for every dungeon in the shared constants table? (e.g., Hallow's End dungeon etc)

DB/Toys/Shadowlands.lua
```diff
@@ -487,7 +487,7 @@ local shadowlandsToys = {
 		},
 		itemId = 187075,
 		chance = 50,
-		requiredDungeons = { [CONSTANTS.LFG_DUNGEON_IDS.TORGHAST_LAYER] = true },
+		requiredDungeons = { [CONSTANTS.LFG_DUNGEON_IDS] = true },
```
Duckwhale, line 490: And this, I believe, inserts the actual table as the key, not the relevant dungeon IDs?

Duckwhale (approved):

PR #662, opened 2023-10-15
wip ui for search bar, no functionality

cyriun:


Duckwhale:
It seems there is a template for search boxes that could be reused:

* https://www.townlong-yak.com/framexml/live/Settings/Blizzard_SettingsPanel.xml#76
* https://www.townlong-yak.com/framexml/live/SharedUIPanelTemplates.xml#1483

I haven't looked into how it works exactly, but the search functionality itself should be doable as well.

As far as I can tell, the `group` table [here](https://github.com/WowRarity/Rarity/blob/master/Modules/Options/Options.lua#L1696) could be filtered to remove all entries that don't match the given search term. Implementing a filter shouldn't be too difficult once the actual location in the code where it should be applied has been identified:

* For simplicity's sake, assume we're only filtering the (localized) item name
* Use Lua's string pattern matching capabilities to check if `item.name` is a match
* If not, remove the item from the table (only the GUI display group, not the actual database/saved variables of course)

Maybe that helps with the rest of this PR? :) Thanks for starting on this task.

rdw-software:
It doesn't seem like there's anyone who currently has the bandwidth to design and implement this feature.

I'll close the PR until that changes. We can use the above as a starting point if/when the issue is revisited :)

Issue #675, opened 2023-10-27
Add Sunreaver Micro-Sentry to the database


Duckwhale:
Source: [Discord](https://discord.com/channels/788119147740790854/788145559654760458/1167477564600746024)
> is there a way to have rarity track attempts to get the Sunreaver Micro-Sentry pet that drops from the Haywire Sunreaver Construct in Isle of Thunder? currently it does not

Seems that this was overlooked, or perhaps deemed not rare enough at the time? I see no reason why it shouldn't be added.

Link: https://www.wowhead.com/item=94124/sunreaver-micro-sentry

Duckwhale:
Resolved via #676.

PR #676, opened 2023-10-31
Add Sunreaver Micro-Sentry to the database (as well as the UiMapID for Isle of Thunder)

cyriun:


cyriun:
from #675

Duckwhale:
Thanks! It looks like you forgot to actually make use of the `UiMapID` for Isle of Thunder, so I've made the necessary changes.

cyriun:
> Thanks! It looks like you forgot to actually make use of the `UiMapID` for Isle of Thunder, so I've made the necessary changes.

rightt, haha, i copied same format as above but its using `uimapid 554` which is timeless isle and not added into `SharedConstants`. Oops

PR #677, opened 2023-11-02
Implement tracking for Brul (pet)

cyriun:
some things to do - possibly, if it even works. 

- [x] shorten long ass sourceText
- [x] give function better name
- [x] fix inevitable conflictions
- [x] test it, didnt have time to wait 30+mins for rare to spawn
- [x] rework so that auras enum can be used in handler

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -27,6 +27,7 @@ local UnitGUID = UnitGUID
 local LoadAddOn = LoadAddOn
 local GetBestMapForUnit = _G.C_Map.GetBestMapForUnit
 local GetMapInfo = _G.C_Map.GetMapInfo
+local IsQuestFlaggedCompleted = C_QuestLog.IsQuestFlaggedCompleted
 local UnitCanAttack = _G.UnitCanAttack
 local UnitIsPlayer = _G.UnitIsPlayer
 local UnitIsDead = _G.UnitIsDead
@@ -1931,6 +1932,10 @@ function R:OnLootReady(event, ...)
 			Rarity:OnDisgustingVatFished()
 		end
 
+		if Rarity.isOpening and Rarity.lastNode == L["Chest of Massive Gains"] then
+			Rarity:OnChestOfMassiveGainsOpened()
+		end
+
 		-- Handle mining Elementium
 		if
 			Rarity.relevantSpells[Rarity.previousSpell] == "Mining"
@@ -2069,7 +2074,29 @@ function R:OnLootReady(event, ...)
 	end
 end
 
-local IsQuestFlaggedCompleted = C_QuestLog.IsQuestFlaggedCompleted
+function Rarity:OnChestOfMassiveGainsOpened()
+	Rarity:Debug("Detected Opening on Chest of Massive Gains")
+
+	local hasOpenedChestToday = IsQuestFlaggedCompleted(75325)
+	if hasOpenedChestToday then
+		Rarity:Debug("Skipping this attempt (loot lockout for Chest of Massive Gains is active)")
+		return
+	end
+
+	local wasRequiredAuraFoundOnPlayer = false
+	AuraUtil.ForEachAura("player", "HELPFUL", nil, function(_, _, _, _, _, _, _, _, _, spellID)
+		if spellID == CONSTANTS.AURAS.ROCKS_ON_THE_ROCKS then
+			wasRequiredAuraFoundOnPlayer = true
+		end
+	end)
+
+	if not wasRequiredAuraFoundOnPlayer then
+		Rarity:Debug(format("Required aura %s NOT found on player", L["Rocks on the Rocks"]))
+		return
+	end
+
+	addAttemptForItem("Brul", "pets")
+end
 
 function Rarity:OnDisgustingVatFished()
 	local hasFishedEmmahThisWeek = IsQuestFlaggedCompleted(75488)
--- a/DB/Pets/Dragonflight.lua
+++ b/DB/Pets/Dragonflight.lua
@@ -1172,6 +1172,23 @@ local dragonflightPets = {
 			{ m = CONSTANTS.UIMAPIDS.THE_FORBIDDEN_REACH, x = 29.2, y = 53.0, n = L["Entrance to Zskera Vaults"] },
 		},
 	},
+	["Brul"] = {
+		cat = CONSTANTS.ITEM_CATEGORIES.DRAGONFLIGHT,
+		type = CONSTANTS.ITEM_TYPES.PET,
+		method = CONSTANTS.DETECTION_METHODS.SPECIAL,
+		name = L["Brul"],
+		spellId = 408110,
+		itemId = 205114,
+		creatureId = 204303,
+		chance = 35,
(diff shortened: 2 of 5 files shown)


cyriun:
this now works, but ideally needs a rework to allow the use of the auras enum instead of the variable

DB/Pets/Dragonflight.lua
```diff
@@ -1172,6 +1172,19 @@ local dragonflightPets = {
...
+		cat = CONSTANTS.ITEM_CATEGORIES.DRAGONFLIGHT,
+		type = CONSTANTS.ITEM_TYPES.PET,
+		method = CONSTANTS.DETECTION_METHODS.SPECIAL,
+		name = L["Brul"],
+		spellId = 408110,
+		itemId = 205114,
+		creatureId = 204303,
+		chance = 35,
+		requiredAura = { [CONSTANTS.AURA.ROCKS_ON_THE_ROCKS] = true },
+		sourceText = L['Looted from "Chest of Massive Gains" which spawns after killing rare - Brullo the Strong. Loot chest while "Rocks on the Rocks" buff is active. Buff is obtained from eating food "Rocks on the Rocks"'],
```
Duckwhale, line 1185: Maybe something like this?

```lua
sourceText = format(L["Use %s before opening %s, which spawns after %s is defeated."], L["Rocks on the Rocks"], L["Chest of Massive Gains"], L["Brullo the Strong"]),
```

I reckon if the instructions are unclear, players will consult wowhead anyway.

Locales.lua
```diff
@@ -2035,6 +2035,10 @@ L["Frozen Coffer"] = true
 L["Obsidian Battle Horn"] = true
 L["Emmah"] = true
 L["Box of Rattling Chains"] = true
+L["Brul"] = true
+L["Chest of Massive Gains"] = true
+L['Looted from "Chest of Massive Gains" which spawns after killing rare - Brullo the Strong. Loot chest while "Rocks on the Rocks" buff is active. Buff is obtained from eating food "Rocks on the Rocks"'] =
```
Duckwhale, line 2040: The chest, item, and NPC names should likely be separate phrases. This allows re-using them without requiring a re-translation, and (more importantly) we can already submit the translated phrases by copy/pasting from wowhead.

Core/EventHandlers.lua
```diff
@@ -2071,6 +2077,25 @@ end
 
 local IsQuestFlaggedCompleted = C_QuestLog.IsQuestFlaggedCompleted
 
+function Rarity:grabAura()
```
Duckwhale, line 2080: This seems somewhat convoluted (and poorly named). I can fix it later, if you prefer?

Core/EventHandlers.lua
```diff
@@ -2071,6 +2077,25 @@ end
...
+	local rotrBuff = 407063 -- Rocks on the Rocks buff
+	local unit = "player"
+	Rarity:Debug("Checking aura on the player character.")
+	local hasOpenedCOMGToday = IsQuestFlaggedCompleted(7325)
+	if not hasOpenedCOMGToday then
```
Duckwhale, line 2085: I would definitely move the lockout check to the front (early exit), similar to how it's done for the Disgusting Vault. That would at least reduce the indentation/nesting level by one and allows mentally "checking off" this case before reading the remaining logic.

Core/EventHandlers.lua
```diff
@@ -2071,6 +2077,25 @@ end
 
 local IsQuestFlaggedCompleted = C_QuestLog.IsQuestFlaggedCompleted
 
+function Rarity:grabAura()
+	local rotrBuff = 407063 -- Rocks on the Rocks buff
```
Duckwhale, line 2081: Didn't you mean to use the shared constant here? At least I assume that's the case since you added one for the aura.

Core/EventHandlers.lua
```diff
@@ -2071,6 +2077,25 @@ end
...
+	local unit = "player"
+	Rarity:Debug("Checking aura on the player character.")
+	local hasOpenedCOMGToday = IsQuestFlaggedCompleted(7325)
+	if not hasOpenedCOMGToday then
+		AuraUtil.ForEachAura(unit, "HELPFUL", nil, function(_, _, _, _, _, _, _, _, _, spellId)
+			if spellId == rotrBuff then
+				addAttemptForItem("Brul", "pets")
+			else
+				Rarity:Debug("Aura Not found on " .. unit)
```
Duckwhale, line 2090: Inconsistent spelling aside, it should probably mention what aura was expected to be found on what unit (especially since there will be other auras once the primal storm PR is merged):

```lua
Rarity:Debug(format("Required aura %s not found on unit %s", L["Rocks on the Rocks"], unit))
```

Not sure what format the other PR will use, but it should probably be consistent.

Duckwhale (changes requested):
Thanks! I'm not sure what the state of the TODOs (checkboxes) is, but I left some comments based on the proposed code changes regardless. Feel free to resolve them if you think they're addressed already, or not worth addressing to begin with.

Core/EventHandlers.lua
cyriun: yeah but i couldnt get it to work, guess ive forgotten how its correctly implemented

DB/Pets/Dragonflight.lua
cyriun: wow, i shouldve tested this regardless but i scrapped something similar bc it gave an error, or warning i guess but it does indeed work

Core/EventHandlers.lua
cyriun: yeah, i guess it can be called something like `OnCOMGOpened` but even then the abbreviation in the function name looks odd. COMG being Chest of Massive Gains that is

Core/EventHandlers.lua
cyriun: yeah or it can be scrapped along with some of the other debug messages, only really needed for testing purposes

Core/EventHandlers.lua
cyriun: super random but somehow i deleted a 5 from the quest id???? no idea how i managed that, it supposed to be 75325

Core/EventHandlers.lua
cyriun: i guess the newer function name is a little better but can be improved depending on how long you want function names to be. For instance you could list the entire chest name `OnChestofMassiveGainsOpened` but would you consider that to be too long

Duckwhale: I don't think `OnChestOfMassiveGainsOpened` is too long. Saving on typing with a one-off function name doesn't seem worth it. FWIW, `COMG` is definitely not something anyone would immediately understand :P

Core/EventHandlers.lua
Duckwhale: You can simply reference `CONSTANTS.AURAS.ROCKS_ON_THE_ROCKS` in this case.

Core/EventHandlers.lua
cyriun: yeah that could also be changed to something like `OpenedMassiveGainsToday`

Duckwhale (approved):

PR #679, opened 2023-11-08
Add Winter Night Dreamsaber to DB

cyriun:
not much info currently but i believe this will need to be updated in the coming days to detect only a higher level Dreamseed Cache. For context depending on how far you progress in the event, i assume spawns different caches with higher level rewards and i would guess the mount chance would require max progress in the event

### DIFF
--- a/DB/SharedConstants.lua
+++ b/DB/SharedConstants.lua
@@ -121,6 +121,7 @@ C.UIMAPIDS = {
 	ZARALEK_CAVERN = 2133,
 	THE_THROUGHWAY = 2165,
 	CROSSROADS_OF_FATE = 2194, -- Dawn of the Infinite (subzone)
+	THE_EMERALD_DREAM = 2200,
 }
 
 -- Types of items
--- a/DB/Toys/Dragonflight.lua
+++ b/DB/Toys/Dragonflight.lua
@@ -429,6 +429,19 @@ local dragonflightToys = {
 		chance = 5,
 		coords = { { m = CONSTANTS.UIMAPIDS.THE_FORBIDDEN_REACH, x = 28.8, y = 52.6 } },
 	},
+	["Molten Lava Ball"] = {
+		cat = CONSTANTS.ITEM_CATEGORIES.DRAGONFLIGHT,
+		type = CONSTANTS.ITEM_TYPES.ITEM,
+		isToy = true,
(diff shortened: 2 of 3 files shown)


Core/EventHandlers.lua
```diff
@@ -1987,6 +1987,20 @@ function R:OnLootReady(event, ...)
 			end
 		end
 
+		--Handling for Winter Night Dreamsaber mount
+		if Rarity.isOpening and Rarity.lastNode == L["Dreamseed Cache"] then
+			Rarity:Debug("Detected Opening on " .. Rarity.lastNode .. " (method = SPECIAL)")
+			local v = self.db.profile.groups.mounts["Reins of the Winter Night Dreamsaber"]
+			if v and type(v) == "table" and v.enabled ~= false then
```
Duckwhale, line 1994: You can (and probably should) use `addAttemptForItem` here, so that future refactoring will be easier.

Core/EventHandlers.lua
```diff
@@ -1987,6 +1987,20 @@ function R:OnLootReady(event, ...)
 			end
 		end
 
+		--Handling for Winter Night Dreamsaber mount
+		if Rarity.isOpening and Rarity.lastNode == L["Dreamseed Cache"] then
+			Rarity:Debug("Detected Opening on " .. Rarity.lastNode .. " (method = SPECIAL)")
```
Duckwhale, line 1992: If this entire block was a separate function (e.g., `OnDreamseedCacheOpened`) it could be tested more easily.

Duckwhale (commented):
Let me know when (if) we should merge this :)

In the meantime, here's some minor comments that might be worth considering.

Core/EventHandlers.lua
cyriun: it only needs to be a pretty basic function also right?

`function Rarity:OnDreamseedCacheOpened()
	addAttemptForItem(dreamseedMounts, "mounts")
end`

assuming dreamseedMounts is a table of the mounts and that works within the function

Duckwhale: I guess? Depends on how complicated the logic has to be for this and the other mounts. (also, may need one call per item)

cyriun: for loop seems to work, but since we now know it drops from only the epic seed its abit irrelevant unless we cannot track it that way

PR #683, opened 2023-11-11
Add Dreamseed Cache mounts to the DB

cyriun:
tried to get this back up quickly after my mistake, almost forgot a couple things so let me know if ive forgotten anything else.


cyriun:
also some things to adjust later on for this will obviously be the gigantic seed tracking and drop chances

Core/EventHandlers.lua
```diff
@@ -2082,5 +2088,22 @@ function Rarity:OnDisgustingVatFished()
 	addAttemptForItem("Emmah", "pets")
 end
 
+local dreamseedMounts = {
+	[L["Reins of the Winter Night Dreamsaber"]] = true,
```
Duckwhale, line 2092: Item names can't be localized here. The database uses English names internally, so on non-English client locales this will not find the item and thus fail to add any attempts.

Core/EventHandlers.lua
```diff
@@ -2082,5 +2088,22 @@ function Rarity:OnDisgustingVatFished()
...
+	[L["Reins of the Snowfluff Dreamtalon"]] = true,
+	[L["Reins of the Evening Sun Dreamsaber"]] = true,
+	[L["Reins of the Blossoming Dreamstag"]] = true,
+	[L["Reins of the Springtide Dreamtalon"]] = true,
+	[L["Reins of the Morning Flourish Dreamsaber"]] = true,
+	[L["Reins of the Rekindled Dreamstag"]] = true,
+}
+
+function Rarity:OnDreamseedCacheOpened()
+	Rarity:Debug("Adding attempts for " .. tostring(dreamseedMounts))
```
Duckwhale, line 2102: That `tostring` will print something like `table: 0x7f9dc7197800`. You probably meant to move this into the loop and print the individual item names?

Locales.lua
```diff
@@ -2039,6 +2039,16 @@ L["Emmah"] = true
...
 L["Sunreaver Micro-Sentry"] = true
 L["Haywire Sunreaver Construct"] = true
+L["Reins of the Winter Night Dreamsaber"] = true
+L["Reins of the Snowfluff Dreamtalon"] = true
+L["Reins of the Evening Sun Dreamsaber"] = true
+L["Reins of the Blossoming Dreamstag"] = true
+L["Reins of the Springtide Dreamtalon"] = true
+L["Reins of the Morning Flourish Dreamsaber"] = true
+L["Reins of the Rekindled Dreamstag"] = true
+L["This item drops from Emerald Bounties that started from Gigantic Seeds, it can also be purchased from a vendor."] =
```
Duckwhale, line 2049: There's already a phrase for items that can be purchased from a vendor. If we use that, it doesn't have to be translated twice. Not sure about the Gigantic Seed notice - if this is something Rarity can detect we might as well remove the note. Otherwise, the explanation isn't clear as to what people should do, so players will have to look it up on wowhead anyway...?

Duckwhale (changes requested):
Thanks! There's a few minor issues, but otherwise this looks like a good start.

I don't know what to do with the Gigantic Seed condition, though. Maybe it will require more changes, but if so we can tackle that problem later (in a separate PR).

Locales.lua
cyriun: yeah pretty much, referring to the to it as epic seed is probably more obvious anyways because i see most referring to it as that

Core/EventHandlers.lua
cyriun: i was actually going to remove that but had forgotten, should i leave it in?

Core/EventHandlers.lua
cyriun: isnt another item using the same format?

Core/EventHandlers.lua
Duckwhale: I'll remove it when rebasing, no problem.

Core/EventHandlers.lua
Duckwhale: Where? I'm only seeing English names used with `addAttemptForItem` - which is what I'd expect.

Note that the object names do need to be localized, but internally the English name is used for the item database.

Duckwhale (approved):

Duckwhale:
Not sure what to do about the cache sizes/game objects, so I'll just merge this as-is and think about it later.

PR #731, opened 2024-07-22
Update for 11.0.2

DakJaniels:
Moved depreciated API calls to the relevant systems and updated to the new [Settings System](https://warcraft.wiki.gg/wiki/Patch_11.0.2/API_changes) for the mini-map icon to open to the settings category directly.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -55,6 +55,11 @@ globals = {
 	"READY_CHECK_WAITING_TEXTURE",
 
 	-- FrameXML (Blizzard interface)
+	"C_AddOns",
+	"C_Container",
+	"C_Item",
+	"Settings",
+
 	"AuraUtil",
 	"BackdropTemplateMixin",
 	"ChatFontNormal",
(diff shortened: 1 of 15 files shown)


Duckwhale:
Thanks! I'll need to read up on the API changes before I can do a proper review, but there's one thing I did notice:

In `AddonCompartment.OnClick` there's some copy/pasted code that still uses `InterfaceOptionsFrame_OpenToCategory`. I assume this would have to be changed as well? Other than that, looks good to me at first glance. The CI failures can be fixed by running [stylua](https://github.com/JohnnyMorganz/StyLua), and adding any new subsystem globals in the `.luacheckrc` file so that luacheck doesn't complain about missing globals.

Feel free to resolve those minor issues if you like, otherwise I'll take care of the rest later.

DakJaniels:
I'll let ya do the stylua stuff. I've not had good experiences with it in the past lol 😆 
I do know it has a GitHub action thing though you could maybe setup if that would be better? 

https://github.com/marketplace/actions/stylua

Duckwhale:
The [Autoformat workflow](https://github.com/WowRarity/Rarity/blob/a03059ee3a9f3973601871e0f15dfa2d2c46a7ae/.github/workflows/check-format.yaml#L24) is already using that GitHub Action :)

I'll leave the review and fixups for tomorrow, when the patch is available on live servers.

DakJaniels:
Was able to test the changes on prepatch on they worked for me.

Issue #506, opened 2022-11-16
Update tooltip scanning logic to use the C_TooltipInfo APIs


Duckwhale:
Leftover from #501  - the old code should still work, so updating it is less urgent (and will require clearing out some technical debt).

filliph:
Is the API change the reason why my tooltips look like this:

![image](https://user-images.githubusercontent.com/9424286/202424409-ff094b93-8246-4f37-a327-ea0cf1a2dcfa.png)

Or should I raise a separate issue?

Duckwhale:
Maybe... probably? I was unable to reproduce that particular issue, which makes me think it's caused by another addon interfering with the tooltip handler that Rarity sets. Can you try disabling all other addons?

d4mich4:
I have the same 
Problem but it seems only to happen on lifing(alife) mobs.

will try disabeling addon to and update with information.

Duckwhale:
I think this is what's happening:

1. Any addon creates a tooltip scanning frame - basically, anything that inherits the ``GameTooltipTemplate``
2. Rarity registers its tooltip handler (just once)
3. Whenever *any* tooltip frame updates, ALL hooks are called (instead of just ``GameTooltip``, as previously/intended)
4. This means Rarity's tooltip handler is called whenever any other addon does "something" with their own tooltips

If my suspicions are correct, then it's due to Blizzard's awkward design, where a "global" registry is used but there aren't really any events to subscribe to, except [TOOLTIP_DATA_UPDATE](https://wowpedia.fandom.com/wiki/TOOLTIP_DATA_UPDATE), which is useless in this context.

Before the patch, whenever ``GameTooltip`` (the standard tooltip frame) was shown, Rarity would add its status line as appropriate. Now it is adding the status line to the standard tooltip once for each tooltip that is triggering the hook via Blizzard's internal ``ProcessTooltipPostCalls`` function. This means we have to check which tooltip causes the activity and exit early if it wasn't the standard tooltip.

I'll test this theory shortly and try out the fix. But I'll add that all of this has actually nothing to do with the original issue, which is referencing the tooltip scanning that Rarity performs to get instance lockouts (IIRC). In theory, it should actually cause the duplicate lines to be added while using that tooltip, as well. Not sure why this isn't the case, though.

Duckwhale:
Unfortunately I still can't replicate the problem, so I'll just push a (presumed) fix and see if it helps.

mdavidson37:
[...]

PR #733, opened 2024-08-06
Improve tooltip performance

filliph:
Improve tooltip performance by using TooltipDataProcessor and reusing its tooltip frame.

I have tested this by linking an item in chat and verifying the tooltip still appears with the correct data:
image

---

The reason why this is important is that using the old `hooksecurefunc(GameTooltip` method will also trigger when sorting your items in your bags when a vendor window is open. The tl;dr version is that every item being sorted will trigger one or more `MERCHANT_UPDATE_*` events, which for some reason also loads the tooltip.

Any addon using `hooksecurefunc(GameTooltip` will therefore have the code running hundreds of times a second, causing a client freeze while the sort operation is ongoing. If enough such add-ons are installed, and a big enough sort operation is performed, the game client may cause Windows to offer to close WoW as it appears to no longer be responding.

---

You can test the performance impact of this by filling a 4x32 inventory with random items, then running this macro:
```
/run for i=1,100,1 do C_Container.PickupContainerItem(math.random(1, 4), math.random(1, 32)); C_Container.PickupContainerItem(math.random(1, 4), math.random(1, 32)); end
```
Then open a vendor and press the sort button in your inventory with the vendor window open.


Duckwhale:
Thanks! Very interesting. I'm not all that familiar with the UI system, so I'll have to set aside some time to refresh my memory.

It might take a couple days for me to get around to reviewing this PR. Feel free to ping me (here or on Discord) in case I forget.

filliph:
I think this change should also fix once-and-for-all this ancient ticket I discovered in my "Participated" inbox: https://github.com/WowRarity/Rarity/issues/506 😄

Core/GUI/GameTooltipHooks.lua
```diff
@@ -538,14 +558,15 @@ hooksecurefunc(GameTooltip, "SetQuestLogItem", function(self, type, index)
...
---        if id then
---            processItem(id)
---        end
---    end
+--	if GetNumQuestLogRewards(questID) > 0 then
+--		local _, _, _, _, _, id = GetQuestLogRewardInfo(1, questID)
+--		if id then
+--			processItem(id)
+--		end
+--	end
```
Duckwhale, line 571: Was the format change done on purpose or did the autoformatter malfunction? (If it's the latter, I should look into why this happened...)

Core/GUI/GameTooltipHooks.lua
```diff
@@ -510,6 +510,26 @@ local function processItem(id)
...
+--- Handles updating an item tooltip to add additional information about attempts
+---@param tooltip GameTooltip
+---@param tooltipData TooltipData
+local function onTooltipSetItem(tooltip, tooltipData)
+	if tooltip ~= _G.GameTooltip and tooltip ~= _G.ItemRefTooltip then
+		return
+	end
+
+	local _, itemLink = tooltip:GetItem()
+	if not itemLink or type(itemLink) ~= "string" then
```
Duckwhale, line 522:
```suggestion
	if type(itemLink) ~= "string" then
```

Unless I'm missing something, the only possible failure case is `itemLink == nil`, but `type(nil) == "nil"`:

image

Core/GUI/GameTooltipHooks.lua
```diff
@@ -510,6 +510,26 @@ local function processItem(id)
...
+		return
+	end
+
+	local id = itemLink:match("item:(%d+):")
```
Duckwhale, line 526: Any reason to worry about this match returning `nil`? Not sure if/how often the item link format has changed.

AFAICT `processItem` dodges that bullet, but maybe an extra `assert` could help avoid a future headache?

Core/GUI/GameTooltipHooks.lua
```diff
@@ -510,6 +510,26 @@ local function processItem(id)
 	end
 end
 
+--- Handles updating an item tooltip to add additional information about attempts
+---@param tooltip GameTooltip
+---@param tooltipData TooltipData
```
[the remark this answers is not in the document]
filliph: That was due to copy/pasting from my own add-on where I use the same hook 😅

Core/GUI/GameTooltipHooks.lua
filliph: Not sure, but does it hurt to have the check regardless? 🤔

Core/GUI/GameTooltipHooks.lua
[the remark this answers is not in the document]
Duckwhale: Fair enough. I'd probably remove it in this case; if we want doc comments for the addon APIs that would warrant discussing a viable approach/determining which APIs to document in a separate issue. The benefit is likely low, though.

Core/GUI/GameTooltipHooks.lua
Duckwhale: No problem; I can remove it later. Generally, whatever stylua outputs is the desired format (to keep things simple).

Core/GUI/GameTooltipHooks.lua
Duckwhale: Technically yes since asserts aren't removed in Lua, but that's irrelevant here. I guess adding one would be great, then.

filliph: I'll let you sort that out then as I've personally not dealt with asserts in Lua so I'd not want to mess it up!

Core/GUI/GameTooltipHooks.lua
```diff
@@ -510,6 +510,23 @@ local function processItem(id)
...
+		return
+	end
+
+	local id = itemLink:match("item:(%d+):")
+	processItem(tonumber(id), tooltip)
+end
+
+_G.TooltipDataProcessor.AddTooltipPostCall(_G.Enum.TooltipDataType.Item, onTooltipSetItem)
+
+--[[
```
Duckwhale, line 529: I just noticed this part - is it safe to remove them? If yes, why comment it out? If no, breaking change?

In other words: Does the `TooltipDataProcessor` still work for all those other scenarios? (I'd have to research it)

Core/GUI/GameTooltipHooks.lua
```diff
@@ -510,6 +510,23 @@ local function processItem(id)
...
+	if tooltip ~= _G.GameTooltip and tooltip ~= _G.ItemRefTooltip then
+		return
+	end
+
+	local _, itemLink = tooltip:GetItem()
+	if type(itemLink) ~= "string" then
+		return
+	end
+
+	local id = itemLink:match("item:(%d+):")
```
Duckwhale, line 523:
```suggestion
	local id = itemLink:match("item:(%d+):")
	assert(id, "Failed to extract item ID from item link (format might have changed?)")
```

Core/GUI/GameTooltipHooks.lua
filliph: I left it commented out so it would be easier to re-add if it turned out to not work for _all_ areas.

I personally have not experienced any ill effects 🤔

Duckwhale: That's what Git is used for :P As far as I understand, this `TooltipDataType` should work for all item tooltips. I suppose we can just remove the comments and see if anyone complains about missing a tooltip somewhere (seems unlikely).

filliph: Cheeky 😂 it's still slightly harder to revert part of a commit! 

But point taken, I can adjust it as soon as I get home 👍

PR #735, opened 2024-08-15
Enable the TOC version check script to process comma-separated lists

Duckwhale:
Piling more hacks onto this sketchy script, it now extracts the highest (latest) version number if there are multiple listed.

I don't expect the approach to be any less brittle, but at least it should again function as an alert in case of outdated TOC versions.


Duckwhale:
CI failure is expected since the latest TOC version wasn't properly detected. No need to create a separate PR for the update.

Issue #740, opened 2024-08-16
Assertion failure in GameTooltipHooks.lua after today's release (r734)


dblanch369:
First time I've seen an error come from Rarity in some time... but after today's update, I randomly got this.... Possibly related to the changed noted in changelog?

1x Rarity/Core/GUI/GameTooltipHooks.lua:524: Failed to extract item ID from item link (format might have changed?)
[string "=[C]"]: in function `assert'
[string "@Rarity/Core/GUI/GameTooltipHooks.lua"]:524: in function <Rarity/Core/GUI/GameTooltipHooks.lua:513>
[string "=(tail call)"]: ?
[string "=[C]"]: in function `securecallfunction'
[string "@Blizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua"]:61: in function <...lizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua:56>
[string "@Blizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua"]:81: in function <...lizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua:77>
[string "@Blizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua"]:117: in function <...lizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua:107>
[string "=[C]"]: in function `SetAttribute'
[string "@Blizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua"]:142: in function <...lizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua:135>
[string "@Blizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua"]:288: in function <...lizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua:238>
[string "=[C]"]: in function `securecallfunction'
[string "@Blizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua"]:235: in function <...lizzard_SharedXMLGame/Tooltip/TooltipDataHandler.lua:234>
[string "=(tail call)"]: ?
[string "=[C]"]: in function `SetBagItem'
[string "@Blizzard_UIPanels_Game/Mainline/ContainerFrame.lua"]:1533: in function `OnUpdate'
[string "@Blizzard_UIPanels_Game/Mainline/ContainerFrame.lua"]:1492: in function `OnEnter'
[...]

PR #741, opened 2024-08-17
fix for error with processing keystones.

DakJaniels:
I have tested this in game with no errors. Let me know if you need me to do anything else for this.

### DIFF
--- a/Core/GUI/GameTooltipHooks.lua
+++ b/Core/GUI/GameTooltipHooks.lua
@@ -511,24 +511,24 @@ local function processItem(id, tooltip)
 end
 
 local function onTooltipSetItem(tooltip, tooltipData)
-    if tooltip ~= _G.GameTooltip and tooltip ~= _G.ItemRefTooltip then
-        return
-    end
+	if tooltip ~= _G.GameTooltip and tooltip ~= _G.ItemRefTooltip then
+		return
+	end
 
-    local name, link = tooltip:GetItem()
-    if type(link) ~= "string" or not name then
-        return
-    end
+	local name, link = tooltip:GetItem()
+	if type(link) ~= "string" or not name then
+		return
+	end
 
-    local id = link:match("item:(%d*)")
-    assert(id, "Failed to extract item ID from item link (format might have changed?)")
-    processItem(tonumber(id), tooltip)
+	local id = link:match("item:(%d*)")
+	assert(id, "Failed to extract item ID from item link (format might have changed?)")
+	processItem(tonumber(id), tooltip)
 end
 
 local function onTooltipSetItemFunction(tooltip, tooltipData)
-    if tooltip == _G.GameTooltip or tooltip == _G.ItemRefTooltip then
-        onTooltipSetItem(tooltip, tooltipData)
-    end
+	if tooltip == _G.GameTooltip or tooltip == _G.ItemRefTooltip then
+		onTooltipSetItem(tooltip, tooltipData)
+	end
 end
 
 _G.TooltipDataProcessor.AddTooltipPostCall(_G.Enum.TooltipDataType.Item, onTooltipSetItemFunction)
--- a/Core/GUI/GameTooltipHooks.lua
+++ b/Core/GUI/GameTooltipHooks.lua
@@ -511,18 +511,24 @@ local function processItem(id, tooltip)
 end
 
 local function onTooltipSetItem(tooltip, tooltipData)
-	if tooltip ~= _G.GameTooltip and tooltip ~= _G.ItemRefTooltip then
-		return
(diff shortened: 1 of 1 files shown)


DakJaniels:
Issue #740

Core/GUI/GameTooltipHooks.lua
```diff
@@ -515,14 +515,20 @@ local function onTooltipSetItem(tooltip, tooltipData)
 		return
 	end
 
-	local _, itemLink = tooltip:GetItem()
-	if type(itemLink) ~= "string" then
+	local name, link = tooltip:GetItem()
+	if type(link) ~= "string" or not name then
```
Duckwhale, line 519: Just to make sure I understand correctly: The failure case here is that `GetItem()` returns `nil, link` where `link` is an actual item link? That seems rather peculiar...

Core/GUI/GameTooltipHooks.lua
```diff
@@ -515,14 +515,20 @@ local function onTooltipSetItem(tooltip, tooltipData)
 		return
 	end
 
-	local _, itemLink = tooltip:GetItem()
-	if type(itemLink) ~= "string" then
+	local name, link = tooltip:GetItem()
```
Duckwhale, line 518: Why rename the second return value? Can it also be some other type of link?

This page suggests that it can't: https://warcraft.wiki.gg/wiki/API_GameTooltip_GetItem

Core/GUI/GameTooltipHooks.lua
```diff
@@ -515,14 +515,20 @@ local function onTooltipSetItem(tooltip, tooltipData)
...
+	if type(link) ~= "string" or not name then
 		return
 	end
 
-	local id = itemLink:match("item:(%d+):")
+	local id = link:match("item:(%d*)")
```
Duckwhale, line 523: I'm afraid I don't understand the reason behind this. Could you please clarify why the pattern should change?

AFAIK an item ID  must always have at least one digit. The problem would be the `keystone` prefix, no?

Core/GUI/GameTooltipHooks.lua
```diff
@@ -515,14 +515,20 @@ local function onTooltipSetItem(tooltip, tooltipData)
...
 	assert(id, "Failed to extract item ID from item link (format might have changed?)")
 	processItem(tonumber(id), tooltip)
 end
 
-_G.TooltipDataProcessor.AddTooltipPostCall(_G.Enum.TooltipDataType.Item, onTooltipSetItem)
+local function onTooltipSetItemFunction(tooltip, tooltipData)
+	if tooltip == _G.GameTooltip or tooltip == _G.ItemRefTooltip then
```
Duckwhale, line 529: I assume this is intended to avoid the case where it's a `keystone` link? If so, isn't that eliminating the need to change the actual tooltip function, since the only type of links it would see are `item` links?

Duckwhale (commented):
Thanks for the quick fix! I'm so far unclear on a few minor things (see review comments).

Core/GUI/GameTooltipHooks.lua
DakJaniels: This function reminded me of another that the game has for the PTR_IssueReporter. So, I just adapted it to Rarity, here is the function I am talking about.

```Lua
function PTR_IssueReporter.SetupItemTooltips()
    local function attachItemTooltip(self)
        local name, link = self:GetItem()
        if (link) and (name) then
            local id = string.match(link, "item:(%d*)")
			-- Professions refactor requires an update to this code in both fetching the current recipe ID and retrieving the
			-- reagent information from the mouse position.
           --if (id == "" or id == "0") and TradeSkillFrame ~= nil and TradeSkillFrame:IsVisible() and GetMouseFocus().reagentIndex then
           --    local selectedRecipe = TradeSkillFrame.RecipeList:GetSelectedRecipeID()
           --    for i = 1, 8 do
           --        if GetMouseFocus().reagentIndex == i then
           --            id = C_TradeSkillUI.GetRecipeReagentItemLink(selectedRecipe, i):match("item:(%d+):") or nil
           --            break
           --        end
           --    end
           --end
            if (id) then
                PTR_IssueReporter.HookIntoTooltip(self, PTR_IssueReporter.TooltipTypes.item, id, name, nil, nil, link)
            end
        end
    end

	local function onTooltipSetItemFunction(tooltip, tooltipData)
		if (tooltip == GameTooltip or tooltip == ItemRefTooltip) then
			attachItemTooltip(tooltip)
		end
	end
	TooltipDataProcessor.AddTooltipPostCall(Enum.TooltipDataType.Item, onTooltipSetItemFunction)
end
```

PR #744, opened 2024-08-19
Fix assertion failure introduced by the recent tooltip changes

DakJaniels:


Duckwhale:
Thanks! Not sure why luacheck doesn't see this assignment. I'll have to look into that when I'm on a different machine.

zeptognome:
the line should be
`
local itemLink = tooltipData.hyperlink
`

Dak added the _

Duckwhale:
Oh, right. I completely missed that - indeed there is only one value now :joy: (Easily fixed, though)

Core/GUI/GameTooltipHooks.lua
```diff
@@ -515,7 +515,7 @@ local function onTooltipSetItem(tooltip, tooltipData)
 		return
 	end
 
-	local _, itemLink = tooltip:GetItem()
+	local _, itemLink = tooltipData.hyperlink
```
Duckwhale, line 518:
```suggestion
	local itemLink = tooltipData.hyperlink
```

Duckwhale:
I didn't encounter any issues in my testing, so presumably this should be fine. If not, can revisit after the next release.

PR #752, opened 2024-08-23
Migrate the Options module from AddOnLoader to LoadOnDemand

DakJaniels:
Removes ancient optional dependency to manage loading the options, blizzard allows addons to set a LoadOnDemand field in the toc for this.

### DIFF
--- a/Modules/Options/Options.lua
+++ b/Modules/Options/Options.lua
@@ -105,9 +105,7 @@ do
 		R.modulesEnabled.options = true
 
 		R:PrepareOptions()
-		if AddonLoader and AddonLoader.RemoveInterfaceOptions then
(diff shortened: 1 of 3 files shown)


Duckwhale:
Thanks! I guess the TOC line here can also be removed?

https://github.com/WowRarity/Rarity/blob/475fb284817c0a2e82601c9d36d3b057f55df361/Rarity.toc#L25

DakJaniels:
> Thanks! I guess the TOC line here can also be removed?
> 
> https://github.com/WowRarity/Rarity/blob/475fb284817c0a2e82601c9d36d3b057f55df361/Rarity.toc#L25

Yes, forgot that one from my local copy. :)

PR #804, opened 2025-01-20
Fixing wrong syntax for utility function FormatTime

Protuhj:
With `PP:FormatTime` all times were being returned as "0:00" since the "self" argument was missing, and all times would appear as `nil` for the `t` argument, causing the 

```lua
if not t then
    return "0:00"
end
``` 
code to always run.

Particularly, this fixes the empty "Time" column in the main display.

image

### DIFF
--- a/Utils/PrettyPrint.lua
+++ b/Utils/PrettyPrint.lua
@@ -46,7 +46,7 @@ end
 
 --- Prints timestamps in a human-readable fashion?
 -- TODO: LuaDoc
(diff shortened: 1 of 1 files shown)


rdw-software:
Thanks! That's indeed a silly oversight, which was apparently introduced in 63fe83e9302f9b152bde57fd5905357faeb147eb.

The CI failures are unrelated (GitHub updated their runner images); I'll fix them shortly.

PR #808, opened 2025-01-27
Add support for Cataclysm Classic servers

aSnejbjerg:
Add support for Rarity to be used in Cataclysm Classic.


rdw-software:
I've haphazardly fixed the CI failures so that a first alpha can be packaged. I'll take a closer look at the changes later.

Thanks again for working on this! Given the scope of the PR, I might pull out a few items if that makes things easier.

aSnejbjerg:
In regards to the Currency info thing. I completely forgot to revert that!
That part of the code is no longer reached in the classic version so it should be reverted :)

rdw-software:
Sorry for the delay! I'm trying to reduce the scope of this PR, but the merge commits prevent rebasing this branch onto main.

I'll see if I can cherry-pick your commits without causing too many conflicts. If all else fails we might have to close this PR and merge the changes piece by piece until everything has been integrated (you'll still be credited as the author either way).

rdw-software:
Just for the record: There's a number of changes that I've made on top of your original branch:

* Added a bunch of (very verbose and not exactly robust) validation tests to make sure no items get lost... hopefully
* Instead of using several nearly-identical TOC files, there is just one for all client versions (easier to maintain)
* The database modules themselves are responsible for detecting when items should be exported or modified
* They do this by using the [expansion levels](https://warcraft.wiki.gg/wiki/LE_EXPANSION) instead of product lines as it should always work, for all WOW projects

There's several concerns that will have to be addressed later on. I'm merging this so that we can get a new release out :)

PR #824, opened 2025-02-28
Migrate Crashin' Thrashin' Battleship attempts on load

rdw-software:
This item had a typo in its name (extra parenthesis). While splitting the holiday events database, the typo was fixed. Unfortunately that means anyone who had attempts recorded under the old name would suddenly find them missing.

The easy way around this would be to just restore the old name and leave it unchanged, forever. But there are some other items that would benefit from a migration system: Firstly, the "Big Love Rocket" is no longer named that, and then there's the manual "purge" command for Island Expedition items that never made it to live. Whether this approach is right, I've no idea.

The migration logic is extremely basic so as to avoid destroying any data. Keeping in mind there are no tests for this yet, it does seem like the safest route. Eventually there may be tests and then this functionality can be extended. For now, I'd rather not mess with the DB internals any more than necessary, especially because many people don't seem to create any backups.

### DIFF
--- a/Core.lua
+++ b/Core.lua
@@ -328,6 +328,7 @@ do
 		if self.db.profile.verifyDatabaseOnLogin then
 			self.Validation:ValidateItemDB()
 		end
+		self.Database:ApplyMigrations()
 	end
 end
 
--- a/Core/Database.lua
+++ b/Core/Database.lua
@@ -1,5 +1,8 @@
 local _, addonTable = ...
 
+local pairs = pairs
+local tostring = tostring
+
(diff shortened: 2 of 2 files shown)


aSnejbjerg (approved):
This seems like a good solution. 
We'll of course have to do this manually for all items that needs to be migrated but that shouldn't happen too often.

rdw-software:
Thanks for the review! I've decided to postpone adding this feature as it would benefit from more refinement (and testing).
For the next release, I'll simply restore the typo. Migrations are still a valuable addition, but they shouldn't be rushed.

PR #856, opened 2025-03-12
Generalize tracking code to support multiple progress bars

### LINKED ISSUE #6: Add support for multiple progress bars
[WowAce Link](https://www.wowace.com/projects/rarity/issues/208)

> The Isle of Time is the one place where 4 things can be farmed at once, currently. Rarity only supports 2 bars. Let's see if we can make it 5 for good measure.

If anything, there shouldn't be a hard limit so Rarity can track as many items as required.
Realistically, this probably won't happen (anytime soon, if ever), but I'm adding it here to consolidate all issues.

aSnejbjerg:
Some cleanup will be required as well as being able to save all the items on the list.


rdw-software:
Sorry for the delay. I've tried to untangle the merge/revert chain as follows:

```sh
git reset --hard 526ca2d9d678d8a513638c361cdee24888d4c1f5
git rebase master
./autoformat.sh
```

For reference, the original commits are saved here: https://github.com/WowRarity/Rarity/tree/backup-pr-856

rdw-software:
The localization phrases for non-English locales are pulled from the WowAce translation system, so they needn't be committed. 
There's also some warnings that you might want to review: You can either install [luacheck](https://github.com/lunarmodules/luacheck) locally or check the CI logs ([example](https://github.com/WowRarity/Rarity/actions/runs/14544055625/job/40806969871?pr=856)).

I'll take a look at the code later, but at first glance it seems like there's a bit more cleanup left to do :)

Core/Tracking.lua
```diff
@@ -102,8 +110,7 @@ end
 -- TODO: What's up with the Hyacinth Macaw? Leaving it at "None" might be a better default?
 function Tracking:FindTrackedItem()
 	self = Rarity
-	Rarity.Tracking:SetTrackedItem(self.db.profile.groups.pets["Parrot Cage (Hyacinth Macaw)"])
-
+	Rarity.Tracking:SetTrackedItem(self.db.profile.groups.pets["None"])
```
rdw-software, line 113: What are the intended semantics of adding "None" here? There's no real item using that key, is there?
Wouldn't `nil` be a better indicator of "nothing is being tracked right now"? If that's in relation to the above TODO comment (which I should've removed long ago - my bad again), it's probably referring to `nil` as well.

Core/Tracking.lua
```diff
@@ -102,8 +110,7 @@ end
 -- TODO: What's up with the Hyacinth Macaw? Leaving it at "None" might be a better default?
```
rdw-software, line 110: This can definitely be removed if you also retire the good old Macaw :) ("None" isn't to be taken literally)

Core/Tracking.lua
```diff
@@ -121,5 +128,94 @@ function Tracking:FindTrackedItem()
 	end
 end
 
+function Tracking:AddTrackedItem(item)
+	Rarity:Debug("AddTrackedItem function called")
```
rdw-software, line 132: That level of tracing seems excessive, no? It should also be possible to use [hooks](https://www.lua.org/manual/5.1/manual.html#pdf-debug.sethook) instead - at least locally.

Core/Tracking.lua
```diff
@@ -121,5 +128,94 @@ function Tracking:FindTrackedItem()
...
+	if not item then -- Make sure the parameter is set.
```
rdw-software, line 133: Comment seems redundant; early exit on `nil` is fairly standard in Lua.

Core/Tracking.lua
```diff
@@ -121,5 +128,94 @@ function Tracking:FindTrackedItem()
...
+		Rarity:Debug("item parameter needs to be set")
+		return
+	end
+	if not item.itemId then -- Make sure the parameter is set.
+		Rarity:Debug("item.itemId parameter needs to be set")
```
rdw-software, line 138: Technically, you'd have to validate every possible parameter to be sure. There's actually some older code that does this but it hasn't been updated in a while. At some point that could be refurbished and then we may `ValidateItem` wherever needed, but that's really not a priority for now. I'd probably just remove this.

Or do you expect items without an ID to be passed into this? I don't think we can use those anyway.

Core/Tracking.lua
```diff
@@ -121,5 +128,94 @@ function Tracking:FindTrackedItem()
...
+		end
+		Rarity:Debug("List now consists of:")
+		for k, v in pairs(Rarity.db.profile.trackedItems.itemList) do
+			local _, trackedItemLink, _, _, _, _, _, _, _, _, _ = GetItemInfo(v.itemId)
+			Rarity:Debug("%s: %s", k, trackedItemLink or "ITEM_NOT_FOUND")
+		end
+	else
+		Rarity:Debug("%s is already on the tracking list", itemLink or "ITEM_LINK_UNAVAILABLE")
+		Rarity.db.profile.trackedItems.itemList[item.itemId] = Tracking:CopyTable(item)
+	end
```
rdw-software, line 166: This part seems somewhat over-complicated, but maybe I misunderstand what you had in mind?
If we have a single list of items, then we can insert or remove via `table.insert` and `table.remove`.
The maximum number of elements can be enforced regardless and you can even use `#` to count.
Rendering them (displaying actual progress bars) would also just be a single loop. Am I missing something?

As for debugging: You can use the table inspector and in the future, we may also hook the progress bar and tracked items up to the (currently still unfinished) debug UI. And the LibBars APIs are library code.

Modules/Options/Options.lua
```diff
@@ -1176,6 +1176,23 @@ function R:PrepareOptions()
 									Rarity.GUI:UpdateText()
 								end,
 							},
+							elements = {
```
rdw-software, line 1179: Does this need a description or will players immediately understand what it's for? (Can't test ingame)

Modules/Options/Options.lua
```diff
@@ -1176,6 +1176,23 @@ function R:PrepareOptions()
...
+								order = newOrder(),
+								type = "range",
+								width = "double",
+								name = L["Max Elements"],
```
rdw-software, line 1183: Something tells me this name might be a little unclear, but I guess we'll have to ask others for feedback.

rdw-software (changes requested):

Core/GUI/DataBrokerDisplay.lua
```diff
@@ -167,7 +167,12 @@ function GUI:UpdateText()
 			end
 		end
 	end
+	ShowTrackedItemList()
+	if true then
+		return
+	end
 
+	--ALL CODE BELOW IS DEPRECATED AND IS RUN IN THE ShowTrackedItemList() FUNCTION INSTEAD
```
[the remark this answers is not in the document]
aSnejbjerg: Yeah, that's a wrong word there :D 
It can indeed be removed. It's no longer used (I honestly forgot to do so)

Core/GUI/DataBrokerDisplay.lua
```diff
@@ -269,5 +274,95 @@ function GUI:UpdateText()
 			self.bar2:SetValue(chance, 100)
 		end
 	end
-	self.Profiling:EndTimer("GUI.UpdateText")
+	self:ProfileStop("UpdateText: %fms")
```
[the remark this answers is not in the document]
aSnejbjerg: I don't see a need for it, so it can be removed. I left it in because I wasn't completely sure about its intended purpose. :)

Core/GUI/DataBrokerDisplay.lua
```diff
@@ -269,5 +274,95 @@ function GUI:UpdateText()
...
+end
+
+function ShowTrackedItemList()
+	--Rarity:Debug("Showing entire tracked item list")
```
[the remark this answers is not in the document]
aSnejbjerg: Nope, let's remove them.

Core/GUI/DataBrokerDisplay.lua
```diff
@@ -269,5 +274,95 @@ function GUI:UpdateText()
...
+	if not trackedItems then
+		return
+	end
+	for key, value in pairs(trackedItems) do
+		local currentItem = value
+		--Rarity:Debug("Current item: %s", key)
+
+		if currentItem == nil or currentItem.itemId == nil then
+			--Rarity:Debug("Current item doesn't exist, ending here")
+			self.barGroup:RemoveBar(key)
```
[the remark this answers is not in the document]
aSnejbjerg: It was at some point Rarity (which is R in the top of this file right now)

Core/Tracking.lua
```diff
@@ -7,12 +7,16 @@ local Tracking = {}
 -- Globals
 local R = Rarity
 -- Locals
+local trackedItemList = {}
+local trackedItemOrder = {}
```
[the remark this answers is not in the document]
aSnejbjerg: When making writing the code (and I'm sure a part of this is my lack of knowledge of lua tables) the trackedItemList wasn't necessarily ordered by the time they were inserted. So I had to make a new list to keep track of the order. If there's a better way I'm all in.

Core/Tracking.lua
```diff
@@ -7,12 +7,16 @@ local Tracking = {}
...
+local trackedItemCount = 0
 local trackedItems = {}
 local lastAttemptItem
 local lastAttemptTime
 -- WOW APIs
 local format = format
 local GetTime = GetTime
+local GetItemInfo = C_Item.GetItemInfo
```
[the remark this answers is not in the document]
aSnejbjerg: If we have that, let me know where to call it and I'll use that instead :D

Core/Tracking.lua
```diff
@@ -53,9 +57,13 @@ function Tracking:Update(item)
 		return
 	end
 	if self.db.profile.trackedItem == item.itemId then
+		if Rarity.db.profile.trackedItems.itemList[item.itemId] then
+			Rarity.db.profile.trackedItems.itemList[item.itemId] = Tracking:CopyTable(item)
```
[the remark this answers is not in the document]
aSnejbjerg: I started by trying to not copying the item over but I very often found myself in an error when loading the game. It kept saying the item doesn't exist.
I'm guessing it's because of some sort of race condition where this part of the code tries to read something before the database files have been read.

Core/Tracking.lua
aSnejbjerg: We could set it to nil, I just read your(?) todo as something that needed to be solved and I was in this part of the code anyways :)

For my sake we can keep it as the macaw and let it be a quirk of the OG dev

Core/Tracking.lua
aSnejbjerg: That should've been removed. I used to to see if the function was called because I got no feedback at some point :D

Core/Tracking.lua
aSnejbjerg: Yeah, let's remove it.

Core/Tracking.lua
aSnejbjerg: All the debugs can probably be removed at this point but as stated in previous comments I had some issues with tables and how to read the data in them :D 
From searching stackOverflow this was the solution I came up with. If we can reduce it to one table that just keeps track of itemId's and pulls data from the DB (without previously mentioned error where it can't find items) I'd greatly prefer that. :)

Core/Tracking.lua
```diff
@@ -121,5 +128,94 @@ function Tracking:FindTrackedItem()
...
+		Rarity:Debug("%s is already on the tracking list", itemLink or "ITEM_LINK_UNAVAILABLE")
+		Rarity.db.profile.trackedItems.itemList[item.itemId] = Tracking:CopyTable(item)
+	end
+end
+
+function Tracking:GetTrackedItemList()
+	return Rarity.db.profile.trackedItems.itemList
+end
+
+function Tracking:CopyTable(tbl)
```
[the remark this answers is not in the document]
aSnejbjerg: Oh so Blizzard has a solution for this? Awesome!
This might not be necessary though if we can move away from having a cached table.

Modules/Options/Options.lua
aSnejbjerg: It's under the category for the tracking bar so I think from context people should understand. We can have a description if needed.

Core/Tracking.lua
```diff
@@ -121,5 +128,94 @@ function Tracking:FindTrackedItem()
...
+			local _, trackedItemLink, _, _, _, _, _, _, _, _, _ = GetItemInfo(v.itemId)
+			Rarity:Debug("%s: %s", k, trackedItemLink or "ITEM_NOT_FOUND")
+		end
+	else
+		Rarity:Debug("%s is already on the tracking list", itemLink or "ITEM_LINK_UNAVAILABLE")
+		Rarity.db.profile.trackedItems.itemList[item.itemId] = Tracking:CopyTable(item)
+	end
+end
+
+function Tracking:GetTrackedItemList()
```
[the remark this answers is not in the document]
rdw-software: Might be due to the asynchronous loading procedure (as discussed on Discord). If the TOC option helps, that's why.

Core/Tracking.lua
```diff
@@ -53,9 +57,13 @@ function Tracking:Update(item)
 		return
 	end
 	if self.db.profile.trackedItem == item.itemId then
+		if Rarity.db.profile.trackedItems.itemList[item.itemId] then
```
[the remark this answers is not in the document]
rdw-software: If caching helps avoid expensive lookups then I'm not complaining - although we might already have this somewhere.

Core/Tracking.lua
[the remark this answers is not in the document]
rdw-software: Worth revisiting after trying out synchronous loading of the saved variables, I suppose.

Core/Tracking.lua
rdw-software: The Lua manual has an example for tracing function calls, and I suspect Blizzard also has some utilities (maybe...).

Core/Tracking.lua
rdw-software: I think Rarity works off the assumption that items are valid once they get past the import stage (for custom items)?
There are a lot of checks in many places, but IMHO validation should be performed once (where needed) only.

Core/Tracking.lua
rdw-software: I haven't spent much thought on how to manage the lookups and deferred loading problems, but in theory one table should be enough. If the problem with DB initialization is the only blocker we can simplify the design... presumably.

PR #872, opened 2025-05-28
Additional sorting optimizations

AySz88:
A few years ago, I contributed some changes (#409 and #411) to sorting to speed up interactions with the main list of items (the "primary tooltip"), focusing on simplification and usage of the built-in library. But zone sorting was still quite slow, and the UI has still been fairly laggy. After accumulating a [few tweaks on my local copy](https://github.com/WowRarity/Rarity/commit/2740df8ee055033af118af00e3ab911330e8c61b) over time, I recently had some inspiration for some [bigger changes to accelerate sorting](https://github.com/WowRarity/Rarity/commit/bae8dd4419c64f75e664db7131d6b0995688d74e).

Using the profiler in Rarity's debug mode, these changes push the sorting times down to around 25ms total worst case on my machine, with options set to show as many items as possible. (This is down from roughly 750ms for sorting by zone and 50ms for other sorts than by name.)

If not for other sources of delay, 25ms would likely be a barely noticeable hitch. Unfortunately there seems to still be quite a lot of time elsewhere (apparently AddGroup?), outside of sorting.

A quick look makes me suspect that further improvements will come from data types that align with the addon's abstractions and usage patterns - for example, an item's "zone" is at different times: the in-game concept of a "zone" as a map area, the list of multiple such zones where the item is found, the string that is the zone's name, what the item row's "zone" entry on the primary tooltip is trying to convey (e.g. not just name, but whether it is in the current zone, or multiple zones), and the formatted string as given to the UI. The UI-formatted string seems canonical at the moment, and is being manipulated and converted ad-hoc. Clearer object models (even MVC?) or abstract data structures could avoid repetitious conversions and the inefficiencies from Lua's very slow string manipulation.

Meanwhile, I hope the sorting itself is getting fast enough to motivate additional work at UI responsiveness.

(As full disclosure, the larger set of changes used assistance from Copilot with Gemini for refactoring and help with a few Lua-specific esoterics.)

### DIFF
--- a/Core/GUI.lua
+++ b/Core/GUI.lua
@@ -51,11 +51,6 @@ local RequestLFDPlayerLockInfo = RequestLFDPlayerLockInfo
 
 -- Addon APIs
 local FormatTime = Rarity.Utils.PrettyPrint.FormatTime
-local sort = Rarity.Utils.Sorting.sort
-local sort_difficulty = Rarity.Utils.Sorting.sort_difficulty
-local sort_category = Rarity.Utils.Sorting.sort_category
-local sort_zone = Rarity.Utils.Sorting.sort_zone
-local sort_progress = Rarity.Utils.Sorting.sort_progress
 local GetDate = Rarity.Utils.Time.GetDate
 local AuctionDB = Rarity.AuctionDB
 
--- a/Core/GUI/MainWindow.lua
+++ b/Core/GUI/MainWindow.lua
@@ -13,7 +13,7 @@ local IsQuestFlaggedCompleted = _G.C_QuestLog.IsQuestFlaggedCompleted
 local C_Covenants = _G.C_Covenants
 
 local FormatTime = Rarity.Utils.PrettyPrint.FormatTime
-local sort2 = Rarity.Utils.Sorting.sort2
+local sort_num = Rarity.Utils.Sorting.sort_num
 local GetDate = Rarity.Utils.Time.GetDate
 local AuctionDB = Rarity.AuctionDB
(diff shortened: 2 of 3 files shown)


AySz88:
So the overall theme is about avoiding work in comparators, since they run much more often than the other code involved in sorting.

The main changes (in chronological order) were:
1. Removing the nil and table type checking from the comparators, since we already ensure that when the table is copied
2. Optimizations to the comparators themselves, so they sort equivalently (or almost so) but are cheaper to evaluate
3. Calculating or generating the data to be compared just once, rather than doing so every time they are needed by a comparator

While the first two parts improved performance marginally, the main step change (particularly for sorting by zone) happened with (3). It's possible (3) can be done without the other changes already included, but honestly it seems odd to make the comparators more expensive again. (Unless something I did in (2) changed ordering in some undesirable way?)

Unfortunately I did not make smaller commits along the way in my local copy of the addon, and (3) effectively became a rewrite of the file. So the first two commits reflect (1 and 2) and (3) respectively.

---

If I understand correctly, I think I agree with you about preferring removal and simplification of code. Personally, I find this style of memoization aesthetically inelegant as an acceleration strategy (thus, the reluctance and gap of a few years!).

However, the large magnitude of improvement suggests that something about the items and functions *outside* Sorting.lua are costly in an unexpected way, and so our preferred solutions involve those larger issues I mentioned...and a far bigger scope.

So the "hidden" cache (and weak references required) are used to maintain an abstraction barrier on the sorting side, to avoid scope creep and keep complexity contained to this file.

PR #879, opened 2025-06-24
11.2 _ first datamine 

Ellezanor:
Added in:
Mounts and Pets from NPC with sources 
Raid toy drop 
New holiday mount - assumed this is the same drop source as the others for this holiday 
New zone + hub (i have added in Tazavesh as Tazavesh Karesh - unsure if there would have been conflict with the SL version if I used Tazavesh, the vieled market)


rdw-software:
I'm not sure which `UiMapID` to use for Manaforge Omega, so I'll just pick the last one from [this list](https://wago.tools/db2/UiMap?filter%5BName_lang%5D=MANAFORGE_OMEGA&page=1).

PR #883, opened 2025-06-28
Add collectibles from A Greedy Emissary (event)

Ellezanor:


Ellezanor:
Ive added the new mounts into a folder to keep the formatting the same - but unsure if this then needs to go elsewhere ?

rdw-software:
Thanks for the PR!

> Ive added the new mounts into a folder to keep the formatting the same - but unsure if this then needs to go elsewhere ?

Creating a new file for TWW holiday mounts is fine. It won't be loaded unless referenced in the [TOC file](https://github.com/WowRarity/Rarity/blob/master/Rarity.toc) though.

Ellezanor:
Will sort the top file later - when the power is back on

PR #884, opened 2025-07-01
Refactor to remove usage of LibBars-1.0

DakJaniels:


rdw-software (commented):
Thanks for the PR!

I didn't have time to fully review the changes. Based on brief testing I've some concerns:

* The icon seems to be glitched with the new bar setup (screenshots posted in Discord)
* The `growUp` setting did not have any discernible effect in the new version (?)
* If we're using a Blizzard template, is it necessary to still maintain that much code?
* Some functions appear to be unused, some for legacy interop, etc.  - can streamline?
* Support for multiple progress bars is parked in #856 - It'd be great to integrate the two

Other than that, the new bar looks pretty neat (IMHO).
I've asked for feedback on Discord, so let's see what people prefer and sort out any technical difficulties in the meantime :)

PR #891, opened 2025-07-06
LibBars-1.0 remove and bar general enhancements

DakJaniels:


DakJaniels:
> Did you mean to close this?

I think it would be better to approach this idea when the other pr's are handled since this would be a breaking change.

rdw-software:
Fair enough. I've created a backup of the PR branch since the original fork is no longer available:

* https://github.com/WowRarity/Rarity/tree/pr-891-backup

Issue #908, opened 2025-09-01
The export UI is broken after 11.2 API changes to Blizzard's StaticPopup dialog


Machou:
Hello,

I have create an item to export it:

![Image](https://github.com/user-attachments/assets/2b5c7b73-bf3e-4b71-97a1-c23052099aa3)

And when i want export it, i have this errors:

<details><summary>Show verbose Lua error details</summary>

```lua
1x Rarity_Options/Options.lua:211: attempt to index field 'editBox' (a nil value)
[Rarity_Options/Options.lua]:211: in function 'OnHide'
[Blizzard_StaticPopup/StaticPopup.lua]:586: in function <Blizzard_StaticPopup/StaticPopup.lua:572>
[C]: in function 'StaticPopup_OnHide'
[Blizzard_StaticPopup_Game/GameDialog.lua]:677: in function <...faceBlizzard_StaticPopup_Game/GameDialog.lua:676>
[C]: in function 'Hide'
[Rarity_Options/Options.lua]:218: in function 'EditBoxOnEscapePressed'
[Blizzard_StaticPopup/SharedTemplates.lua]:58: in function <...faceBlizzard_StaticPopup/SharedTemplates.lua:55>

Locals:
self = StaticPopup1 {
 TopSpacer = FontString {
 }
 dirty = false
 visibleButtons = <table> {
 }
 CoverFrame = Frame {
 }
 minimumWidth = 320
 previousRegionKey = "ButtonContainer"
 widthPadding = 0
 ProgressBarSpacer = Texture {
 }
 exclusive = 1
 DarkOverlay = Frame {
 }
 timeleft = 0
 ButtonContainer = Frame {
 }
 dialogInfo = <table> {
 }
 which = "RARITY_OPTIONS_ALERT_WITH_COPY"
 hideOnEscape = 1
 ItemFrame = Frame {
 }
 MoneyInputFrame = StaticPopup1MoneyInputFrame {
 }
 AlertIcon = Texture {
 }
 ProgressBarFill = Texture {
 }
 ExtraButton = StaticPopup1ExtraButton {
 }
 Buttons = <table> {
 }
 BG = Frame {
 }
 EditBox = StaticPopup1EditBox {
 }
 ProgressBarBorder = Texture {
 }
 Dropdown = Button {
 }
 SubText = FontString {
 }
 heightPadding = 16
 Spinner = Frame {
 }
 numButtons = 2
 MoneyFrame = StaticPopup1MoneyFrame {
 }
[...]

PR #909, opened 2025-09-02
fix: resolves issue #908

DakJaniels:


rdw-software:
(Side note: If you create a named topic branch, there won't be conflicts when checking out your changes locally 😅 )

PR #925, opened 2025-10-20
Set up automated checks for outdated WowAce translations

rdw-software:
This is only the first step needed to automatically handle localization updates.
I don't have time to build out the rest right now, but this initial script will:

* Download the current set of localization phrases from CurseForge (using the web API)
* Compare the `enUS` table against the one found in the source code itself
* Error if any phrases have been added to the addon which aren't present on WowAce yet
* Error if the WowAce system includes phrases which are no longer used by the addon

Obviously an API token must be provided to actually run the script, though the web export should work without one.

Effectively, this is just an automated reminder to manually import the table using WowAce's import tool when needed.
While the workflow is enabled, it should no longer be possible to publish releases without also updating the localization.
The approach has similar limitations as the existing TOC version check (requires manual steps , specific CLI tools, etc.)

In the future, it would be useful to automatically upload the phrases, or possibly mirror them in the source directly.
One could even automatically check that all phrases have been added; I used to have a script for that, but no idea where.
Namespaces would also need to be handled for the TOC phrases to be covered, but they rarely/never change anyway.

### DIFF
--- a/.github/workflows/check-wowace-locales.yml
+++ b/.github/workflows/check-wowace-locales.yml
@@ -1,0 +1,23 @@
+name: WowAce Localization
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches: [master]
+
+jobs:
+  toc:
+    name: Check localization phrases
+    runs-on: macos-latest
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v4
+
+      - name: Install Lua runtime # May be overkill, but removes the need for complex shell scripting
+        uses: evo-lua/evo-setup-action@main
+        with:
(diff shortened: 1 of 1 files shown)


rdw-software:
After manually updating `enUS`, the WowAce localization system should now be in sync with what's being stored on GitHub. Since there's no way to see what WowAce will do when telling it to delete missing phrases, I've stored a backup snapshot here:

* https://github.com/WowRarity/Rarity/tree/backup-localization-exports

FWIW: It doesn't look like anything of value was deleted; the unused phrases were either duplicates that had typos, removed debug features, or obsolete stuff like the LFG update check box (which was killed by Blizzard back in the OQueue days).

PR #930, opened 2025-12-16
UI: Show known status and pickpocket requirements in the same line in tooltip

lostindark:
Before:
image

After:
image


rdw-software:
That's a good point. Let me finish the unit tooltip cleanup first and then we can either get this integrated, or find a different solution (that still fixes the UI problem for your locale). It'll probably take a while, so I'll leave this PR open as a reminder.

PR #941, opened 2026-02-04
WoW 12.0.0 (Midnight) Compatibility

Amadeus-:
### Summary

This patch adds compatibility with WoW 12.0.0 (Midnight), which introduced two major breaking changes for addons:

1. **COMBAT_LOG_EVENT_UNFILTERED is blocked** - Addons can no longer register for CLEU; doing so triggers `ADDON_ACTION_FORBIDDEN`. Boss tracking on 12.0.0+ now relies entirely on the existing `ENCOUNTER_END` handler, which was already registered and functional.
2. **Secret Values system** - During combat, many `Unit*` APIs return opaque "secret values" that cannot be used in string operations, comparisons, or passed as arguments to other API calls. Attempting to do so raises a runtime error.

All changes are backward-compatible with pre-12.0 clients via version-gating and nil-guarded fallbacks.

#### TOC Files
- `Rarity.toc`, `Rarity_Options.toc`: Added `120000, 120001` to `## Interface` and `## X-Min-Interface`

#### API Compatibility Wrappers

| File | API | Fallback |
|------|-----|----------|
| `Core.lua` | `CombatLogGetCurrentEventInfo` | `C_CombatLog.GetCurrentEventInfo` |
| `Core.lua` | `IsSpellKnown` | `C_SpellBook.IsSpellInSpellBook` |
| `Core.lua` | `COMBATLOG_OBJECT_AFFILIATION_*` | Three-tier: global -> `Enum.CombatLogObject.*` -> hardcoded bitmask |
| `EventHandlers.lua` | `CombatLogGetCurrentEventInfo` | `C_CombatLog.GetCurrentEventInfo` |
| `EventHandlers.lua` | `IsSpellKnown` | `C_SpellBook.IsSpellInSpellBook` |
| `EventHandlers.lua` | `GetNumArchaeologyRaces` | Returns 0 if nil |
| `Collections.lua` | `GetNumArchaeologyRaces` | Returns 0 if nil |

#### Combat Log Event Handling (EventHandlers.lua)
- CLEU registration is version-gated at line 75: `if interfaceVersion < 120000 then` registers CLEU; otherwise the event is simply not registered
- On 12.0.0+, boss kill tracking relies on the existing `ENCOUNTER_END` handler which uses a lookup table (`encounterLUT`) to map encounter IDs to tracked items
- No new UNIT_DIED/PARTY_KILL handlers were needed; the existing ENCOUNTER_END coverage is sufficient for raid/dungeon bosses

#### Secret Value Guards

| File | Location/Function | What's Protected |
|------|-------------------|------------------|
| `Core.lua` | `GetNPCIDFromGUID()` | `guid` parameter before `strsplit()` |
| `Core.lua` | `CheckNpcInterest()` | `guid` parameter before `type(guid)` check |
| `EventHandlers.lua` | `OnMouseOver()` | `UnitGUID("mouseover")` return value |
| `EventHandlers.lua` | `OnCursorChanged()` | `tooltipLeftText1:GetText()` return value |
| `EventHandlers.lua` | `OnLootReady()` (legacy section) | `UnitGUID("target")` and `UnitName("target")` |
| `EventHandlers.lua` | `OnLootReady()` (NPC loop) | Each `guid` from `GetLootSourceInfo()` before `type()` call |
| `GameTooltipHooks.lua` | `onTooltipSetUnit()` | `unit` token from `GetUnit()` |
| `GameTooltipHooks.lua` | `onTooltipSetUnit()` | `UnitGUID(unit)` return value |

#### Bug Fixes

**Pre-existing Bug (EventHandlers.lua)**
- Line 2053: Fixed `UnitClassification()` call that was incorrectly passing a GUID instead of the `"target"` unit token

**Announcements.lua - Icon Display Fix**
- **Problem**: Chat announcement messages inconsistently showed item icons (sometimes icon appeared, sometimes not)
- **Cause**: Raw `C_Item.GetItemInfo()` returns nil for uncached items, and there was no fallback
- **Fix applied**:
  1. Changed from raw `C_Item.GetItemInfo()` to Rarity's cached `R:GetItemInfo()` wrapper (takes advantage of item priming on startup)
  2. Added fallback to Rarity's default icon (`spell_nature_forceofnature`) if texture is still nil
- **Result**: Announcement messages now consistently display an icon

#### Version-Gated API Fallbacks (Collections.lua)

Restored backward-compatible fallbacks using API existence checks (not hard-coded version numbers) to support all WoW versions:

**Mount API Fallbacks:**
- Modern API (WoD+): Uses `C_MountJournal.GetMountIDs()` + `GetMountInfoByID()`
- Legacy API (pre-WoD): Falls back to `C_MountJournal.GetNumMounts()` + `GetMountInfo()`
- Detection: `if C_MountJournal.GetMountIDs then ... else ... end`

**Companion System Fallbacks:**
- Added `GetNumCompanions`/`GetCompanionInfo` upvalues
- Added companion scanning loop for Classic Era (pre-MoP)
- Detection: `if GetNumCompanions then ... end` (only runs if API exists)

**Pet Journal Filter Fallbacks:**
- Modern API (Legion 7.0+): `SetFilterChecked()`, `SetAllPetTypesChecked()`, `SetAllPetSourcesChecked()`
- Legacy API (pre-Legion): `SetFlagFilter()`, `AddAllPetTypesFilter()`, `AddAllPetSourcesFilter()`
- Detection: `if C_PetJournal.SetFilterChecked then ... else ... end`

**Bug Fixes Retained:**
- `GetActiveArtifactByRace` nil guard still present
- Added nil check for `name` in archaeology scanning loop

#### UI Fixes
- `MainWindow.lua`: `CombatText_AddMessage` (removed in 12.0) replaced with nil-guarded fallback to `UIErrorsFrame:AddMessage`
- `FauxAchievementPopup.lua`: Added nil guard for `ActionStatus` frame reference
- `EventHandlers.lua`: Wrapped `GetLootInfo()` return in nil check

### Backward Compatibility

All changes are safe for pre-12.0 versions:

1. **API wrappers** use the `OldAPI or NewAPI` pattern, which evaluates to the old API when it exists:
   ```lua
   local CombatLogGetCurrentEventInfo = CombatLogGetCurrentEventInfo or (C_CombatLog and C_CombatLog.GetCurrentEventInfo)
   ```

2. **Secret value guards** use `if issecretvalue and issecretvalue(val)` which short-circuits safely when `issecretvalue` is nil (pre-12.0):
   ```lua
   if issecretvalue and issecretvalue(guid) then return end
   ```

3. **Version-gated CLEU registration** ensures pre-12.0 clients still register the event:
   ```lua
   local _, _, _, interfaceVersion = GetBuildInfo()
   if interfaceVersion < 120000 then
       self:RegisterEvent("COMBAT_LOG_EVENT_UNFILTERED", "OnCombat")
   end
   ```

4. **Collections.lua API existence checks** support all WoW flavors without hard-coded version numbers:
   - **Classic Era (11508)**: Uses Companion system (`GetNumCompanions`/`GetCompanionInfo`)
   - **Classic Cataclysm (50503)**: Uses appropriate APIs for that expansion (legacy mount/pet APIs)
   - **Retail 12.0.0+**: Uses modern namespaced APIs (`GetMountIDs`, `SetFilterChecked`, etc.)

   Example pattern:
   ```lua
   if C_MountJournal.GetMountIDs then
       -- Modern API (WoD+)
       for _, mountID in ipairs(C_MountJournal.GetMountIDs()) do
           local name, id = C_MountJournal.GetMountInfoByID(mountID)
           ...
       end
   else
       -- Legacy API (pre-WoD)
       for i = 1, C_MountJournal.GetNumMounts() do
           local name, id = C_MountJournal.GetMountInfo(i)
           ...
       end
   end
   ```

### DIFF
--- a/Core/EventHandlers.lua
+++ b/Core/EventHandlers.lua
@@ -1144,7 +1144,7 @@ function R:ProcessCollectionItemSingle(collectionItem, itemID)
 	local item = Rarity.items[itemID]
 
 	if
-		collectionItem.enabled
+		collectionItem.enabled ~= false
 		and (
 			collectionItem.collectedItemId == item.collectedItemId
 			or table_contains(item.collectedItemId, collectionItem.collectedItemId)


rdw-software:
Thank you! I don't have the time to read through all these details right now, but I'll take a look tomorrow if I can.

(At first glance: Probably need to split this PR? Also, is this an AI summary? No offense, but should clarify use upfront)

Core/EventHandlers.lua
Amadeus-: Nasty little bug ...easy to miss, especially for those of us who are primarily C++ coders ...lol.

Core/EventHandlers.lua
```diff
@@ -1119,7 +1144,7 @@ function R:ProcessCollectionItemSingle(collectionItem, itemID)
 	local item = Rarity.items[itemID]
 
 	if
-		collectionItem.enabled
+		collectionItem.enabled ~= false
```
Amadeus-: This is a nasty little bug, easy to miss (especially for me as primarily a C++ coder.)

Core/EventHandlers.lua
```diff
@@ -64,7 +69,12 @@ function EventHandlers:Register()
 	self:RegisterEvent("LOOT_READY", "OnLootReady")
 	self:RegisterEvent("CURRENCY_DISPLAY_UPDATE", "OnCurrencyUpdate")
 	self:RegisterEvent("RESEARCH_ARTIFACT_COMPLETE", "OnResearchArtifactComplete")
-	self:RegisterEvent("COMBAT_LOG_EVENT_UNFILTERED", "OnCombat") -- Used to detect boss kills that we didn't solo
+	-- Version-gate combat log event registration for 12.0.0 compatibility
+	-- On 12.0.0+, boss tracking relies on ENCOUNTER_END which is already registered above
```
rdw-software: TBD: Will that automatically pick up the items that use BOSS currently? (ENCOUNTER_END isn't yet a predefined method. It likely should be, but that will require a fallback and proper testing)

PR #943, opened 2026-02-07
Fix disableForClass in ProcessContainerItems function

AlexAndr041:
Add disableForClass check in ProcessContainerItems function


AlexAndr041:
Hi. Yes. Any containers with armor transmog. I used it to track attempts
for drop from 235054 Pristine Mysterious Satchel
<https://www.wowhead.com/item=235054/pristine-mysterious-satchel>

PR #982, opened 2026-09-19
fix(detection): tolerate a calendar day the client returns as secret

Haembina:
`C_Calendar.GetDayEvent()` can return a table whose every field is a secret value — `calendarType`, `eventID`, `title`, all of them. `R:ScanCalendar` compares `calendarEvent.calendarType == "HOLIDAY"`, and comparing a secret string from tainted code throws:

```
Core/Detection.lua:71: attempt to compare field 'calendarType' (a secret string value, while execution tainted by 'Rarity')
```

The next line would fail the same way, using the secret `eventID` as a table key.

The loop now skips an event it cannot read and sets `Rarity.calendarIsSecret`. `HolidayEvents.IsItemAvailableToday()` reads that flag and returns **true**.

That direction is deliberate: the flag only drives the `Unavailable` label. Wrongly showing a holiday item as trackable costs nothing; wrongly hiding one hides it for a year.

The `assert` moves inside the readable branch — it states a claim about `eventID` that no longer survives a secret one.


Roguinaa:
This addon is not currently supported. Check https://www.curseforge.com/wow/addons/rarity-collectors-edition - Previous devs of Rarity are working on it.

